PR #6, opened 2020-03-23
Add activeAtTime to relationships filter

andreavs:


cognite/experimental/_api/relationships.py
```diff
@@ -187,6 +191,7 @@ def list(
         created_time: Dict[str, Any] = None,
         data_set: Optional[Union[str, List[str]]] = None,
         relationship_type: Optional[Union[str, List[str]]] = None,
+        active_at_time: int = None,
```
sanderland: docstring

cognite/experimental/_api/relationships.py
```diff
@@ -116,6 +119,7 @@ def __call__(
             last_updated_time=last_updated_time,
             created_time=created_time,
             data_set=data_set,
+            active_at_time=active_at_time,
```
sanderland: docstring

andreavs:
Testing? I thought this was ~america~ python. Expanded on the a bit on the general list test and added a docstring. I might have missed some standard methods on the test, like i could not find a built-in way to translate camelCase <-> snake_case for the request (even though i was able to find one for response :thinking: )

cognite/experimental/_api/relationships.py
sanderland: still missing docs in call

tests/tests_unit/test_api/test_relationships.py
```diff
@@ -167,13 +170,43 @@ def test_delete_multiple(self, mock_rel_response):
...
+        """
+        Checks that all (non deprecated) fields in the list method are surfaced in the api call
+        """
+
+        def _generate_random_string():
+            string_length = random.randint(5, 10)
+            letters = string.ascii_letters
+            return "".join(random.choice(letters) for i in range(string_length))
+
+        def _to_snake_case(camel_case_string):
```
sanderland: this is in utils

tests/tests_unit/test_api/test_relationships.py
```diff
@@ -167,13 +170,43 @@ def test_delete_multiple(self, mock_rel_response):
...
-            "filter": {"sources": [{"resource": "asset"}], "relationshipTypes": ["belongs_to"]},
-            "limit": 25,
-            "cursor": None,
-        } == jsgz_load(mock_rel_response.calls[0].request.body)
+    def test_advanced_list_fields(self, mock_rel_response):
+        """
+        Checks that all (non deprecated) fields in the list method are surfaced in the api call
+        """
+
+        def _generate_random_string():
```
sanderland: this is also in utils

tests/tests_unit/test_api/test_relationships.py
```diff
@@ -167,13 +170,43 @@ def test_delete_multiple(self, mock_rel_response):
...
+        # TODO: some fields are skipped here because there some deprecated fields are unified with new field, so that
+        # there are inconsistencies in the naming in the api vs the sdk (e.g. there is a relationshipTypes field) in
+        # the api, but no relationship_types in the sdk. If we drop support for the deprecated fields we should be
+        # able to cover all fields.
+        input = {
+            "active_at_time": random.randint(0, 10000),
+            "sources": [{"resource": "asset", "resource_id": _generate_random_string()}],
+            "targets": [{"resource": "asset", "resource_id": _generate_random_string()}],
+            # "relationship_type": ["flows_to"],
+            "created_time": {"min": random.randint(0, 10000), "max": random.randint(10001, 20000)},
```
sanderland: we typically just use 0/1/'a' and such instead of random values

andreavs: but i) aren't random values objectively better? and ii) do we have to conform with the regular sdk on this? Not saying we need to redo every test right away, but i also don't like replacing code with worse code

sanderland: not saying you need to change this, just that the bar is lower :)

PR #13, opened 2020-04-01
[CDF-3672] Add FunctionsAPI.list()

HMEiding:
This PR adds `FunctionsAPI` to the `CogniteClient`. For now this API only has a single method `list()`, with which one can list functions by doing:
```
c = CogniteClient()
functions = c.functions.list()
```
`functions` will then be a list of `Function` objects, and each such object has the expected properties:
```
function1 = functions[0]
print(function1)
```
gives
```
{
    "id": 8235031195001950,
    "name": "sleep",
    "description": "",
    "owner": "",
    "status": "Ready",
    "file_id": 1355689374483250,
    "created_time": "2020-03-29 09:58:22",
    "secrets": {}
}
```
Each of the properties above are accessed as attributes like so:
```
print(function1.name)
```
gives
```
"sleep"
```

Note that values returned as `null` by the `/functions` API, are not printed. This is intentional across all parts of the SDK, so if we want such values to be displayed, we should return for example an empty string for such fields.

### DIFF
--- a/cognite/experimental/data_classes/functions.py
+++ b/cognite/experimental/data_classes/functions.py
@@ -15,7 +15,7 @@ class Function(CogniteResource):
         filed_id (int): File id of the code represented by this object.
         created_time (int): Created time in UNIX.
         api_key (str): Api key attached to the function.
-        secrets (Dict): Secrets attached to the function.
+        secrets (Dict[str, str]): Secrets attached to the function ((key, value) pairs).
         cognite_client (CogniteClient): An optional CogniteClient to associate with this data class.
     """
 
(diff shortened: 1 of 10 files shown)


HMEiding:
Integration tests will have to wait until we have added a `create()` method.

cognite/experimental/data_classes/functions.py
```diff
@@ -0,0 +1,51 @@
+from typing import Dict
+
+from cognite.client.data_classes._base import CogniteResource, CogniteResourceList
+
+
+class Function(CogniteResource):
```
sanderland, line 6: did you type this in manually rather than autogenerate?

HMEiding: Yes, since we don't yet have our spec in service-contracts yet, that's how I had to do it. But an autogeneration should in principle give the same.

PR #14, opened 2020-04-02
[CDF-3676] Add FunctionsAPI.create()

HMEiding:
This PR adds method `FunctionsAPI.create()` which can be used to create a function in two ways:
1) By using the `folder` argument which is the relative path to the folder containing the function source code (i.e. handler.py):
```
client = CogniteClient()

func = client.functions.create(name="myfunction", path="path/to/code")
```

2) By using the `file_id` argument if the source code is already uploaded to the files-API:
```
func = client.functions.create(name="myfunction", file_id=1234)
```
Arguments `name` and exactly one of `folder` or `file_id` are required. Optional arguments are: `external_id`, `description`, `owner`, `api_key`, `secrets`.

The method `client.functions.create()` returns an object of type `Function`:
```
{
    "id": 7344064945410280,
    "name": "henkistest",
    "description": "",
    "owner": "",
    "status": "Queued",
    "file_id": 8560551814397155,
    "created_time": "2020-04-02 22:06:18",
    "secrets": {}
}
```

### DIFF
--- a/cognite/experimental/_api/functions.py
+++ b/cognite/experimental/_api/functions.py
@@ -23,7 +23,7 @@ class FunctionsAPI(APIClient):
 
         Args:
             name (str):                 Name of function
-            folder (str):               Path to folder (relative to current working directory) where the function source code is located
+            folder (str):               Path to folder where the function source code is located
             external_id (str):          External id of the function
             description (str):          Description of the function
             owner (str):                Owner of the function
@@ -76,7 +76,7 @@ class FunctionsAPI(APIClient):
                     zf.write(os.path.join(root, filename))
             zf.close()
 
-            file = self._cognite_client.files.upload(zip_path, name=name)
+            file = self._cognite_client.files.upload(zip_path, name=f"{name}.zip")
 
         os.chdir(current_dir)
 
--- a/cognite/experimental/_api/functions.py
+++ b/cognite/experimental/_api/functions.py
@@ -1,10 +1,58 @@
+import os
+from tempfile import TemporaryDirectory
 from typing import Any, Dict, List
+from zipfile import ZipFile
 
 from cognite.client._api_client import APIClient
-from cognite.experimental.data_classes import FunctionList
+from cognite.experimental.data_classes import Function, FunctionList
 
 
 class FunctionsAPI(APIClient):
+    def create(
+        self,
+        name: str,
+        folder: str = None,
+        file_id: int = None,
+        external_id: str = None,
+        description: str = "",
+        owner: str = "",
+        api_key: str = None,
+        secrets: Dict = None,
+    ) -> Function:
+        """Creates a new function from source code located in folder
+
+        Args:
+            name (str):                 Name of function
+            folder (str):               Path to folder (relative to current working directory) where the function source code is located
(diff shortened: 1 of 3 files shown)


HMEiding:
Integration tests will have to wait until the `delete()` method is added.

cognite/experimental/_api/functions.py
```diff
@@ -1,10 +1,58 @@
...
+        description: str = "",
+        owner: str = "",
+        api_key: str = None,
+        secrets: Dict = None,
+    ) -> Function:
+        """Creates a new function from source code located in folder
+
+        Args:
+            name (str):                 Name of function
+            folder (str):               Path to folder (relative to current working directory) where the function source code is located
```
andeplane, line 26: We don't need to specify `relative to current working directory` as it also works with absolute paths.

cognite/experimental/_api/functions.py
```diff
@@ -14,3 +62,22 @@ def list(self) -> FunctionList:
...
+        with TemporaryDirectory() as tmpdir:
+            zip_path = os.path.join(tmpdir, "function.zip")
+            zf = ZipFile(zip_path, "w")
+            for root, dirs, files in os.walk("."):
+                zf.write(root)
+                for filename in files:
+                    zf.write(os.path.join(root, filename))
+            zf.close()
+
+            file = self._cognite_client.files.upload(zip_path, name=name)
```
andeplane, line 79:
```suggestion
            file = self._cognite_client.files.upload(zip_path, name=f"{name}.zip")
```

cognite/experimental/_api/functions.py
HMEiding: I guess you're correct.

cognite/experimental/_api/functions.py
```diff
@@ -1,10 +1,58 @@
...
+            raise TypeError("Exactly one of the arguments `path` and `file_id` is required, but both were given.")
+        if not folder and not file_id:
+            raise TypeError("Exactly one of the arguments `path` and `file_id` is required, but none were given.")
+
+        if not file_id:
+            file_id = self._zip_and_upload_folder(folder, name)
+
+        url = "/functions"
+        function = {"name": name, "description": description, "owner": owner, "fileId": file_id}
+        if external_id:
```
HMEiding, line 46: The special handling of `externalID`, `apiKey` and `secrets` is because they can't be `None`: They have to be `str`, `str` and `Dict`. An option would be to have default args different from `None` for these (like `""`), but since we don't have the key `missing=True` in the API schema, my understanding is that we don't want e.g. `externalID=""` if the user does not provide this argument. Agree @andeplane ?

sanderland:
bump version?
shall we start a changelog?

HMEiding:
> bump version?
> shall we start a changelog?

Yes, a changelog is good idea. But I think we can wait with the version bumping until we have all the functions-methods in.

andeplane:
Let's wait with bumping until we have all of the features in there!

cognite/experimental/_api/functions.py
andeplane: Yep. Let's patch the API later, but allow this for now.

PR #17, opened 2020-04-06
[CDF-3678] [CDF-3680] Add FunctionsAPI.call()

HMEiding:
This PR adds `FunctionsAPI.call()`, which is used to call functions by `id` or `external_id`. It is used as follows:
```
c = CogniteClient()
call = c.functions.call(id=4172094376815850)
```
Where a `Completed` call looks like:
```
print(call)
{
    "id": 4172094376815850,
    "start_time": "2020-04-06 07:52:03",
    "end_time": "2020-04-06 07:52:04",
    "response": {
        "assetId": 1234
    },
    "status": "Completed"
}
```
Alternatively, you can make the function call from a `Function` object directly:
```
func = c.functions.retrieve(id=4172094376815850)
call = func.call()
```

Optional arguments are `data`, which needs to be JSON serializable, and `asynchronous` (defaults to `False`) which lets you call the function asynchronously.

### DIFF
--- a/cognite/experimental/_api/functions.py
+++ b/cognite/experimental/_api/functions.py
@@ -166,7 +166,11 @@ class FunctionsAPI(APIClient):
         utils._auxiliary.assert_type(external_ids, "external_id", [List], allow_none=True)
         return self._retrieve_multiple(ids=ids, external_ids=external_ids, wrap_ids=True)
 
-    def call(self, id: int, data=None, asyncronous: bool = False) -> FunctionCall:
+    def call(self, id: int = None, external_id: str = None, data=None, asyncronous: bool = False) -> FunctionCall:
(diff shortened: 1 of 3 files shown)


cognite/experimental/_api/functions.py
```diff
@@ -166,6 +166,44 @@ def retrieve_multiple(
         utils._auxiliary.assert_type(external_ids, "external_id", [List], allow_none=True)
         return self._retrieve_multiple(ids=ids, external_ids=external_ids, wrap_ids=True)
 
+    def call(self, id: int = None, external_id: str = None, data=None, asynchronous: bool = False) -> FunctionCall:
```
sanderland, line 169: for the future we should align the async call and what the contextualization api does

PR #18, opened 2020-04-06
Restructure FunctionsAPI

HMEiding:
This PR restructures the FunctionsAPI by adding an additional level in the hierarchy. Previously, functions were retrieved by:
```
c = CogniteClient()
func = c.functions.retrieve(id=123)
```
whereas now, they are retrieved by:
```
func = c.functions.functions.retrieve(id=123)
```
The reason for this restructuring is the anticipation of `calls` and `schedules`, so that one for example will list calls by doing:
```
calls = c.functions.calls.list(function_id=123)
```


sanderland:
please make it
func = c.functions.retrieve(id=123)
call = c.functions.calls.retrieve(id=123)
this would mirror e.g. sequences

HMEiding:
> please make it
> func = c.functions.retrieve(id=123)
> call = c.functions.calls.retrieve(id=123)
> this would mirror e.g. sequences

I thought about that possibility as well. However, three_d and model_hosting has the current setup, so I guess there's no hope to achieve consistency anyhow. But which of the two feels more natural? I don't have any strong feelings one way or the other. However, if we for example want to list all schedules across all functions, then it more natural to have functions and schedules side by side. What do you think @andeplane ?

sanderland:
do we have c.model_hosting.model_hosting or c.three_d.three_d ? I'm pretty sure we don't, Hence no c.functions.functions

HMEiding:
> do we have c.model_hosting.model_hosting or c.three_d.three_d ? I'm pretty sure we don't, Hence no c.functions.functions

For model hosting for example, we a have all parts of the API side by side:
```
mlh = c.model_hosting
models = mlh.models.list_models()
versions = mlh.versions.list_versiosn()
schedules = mlh.schedules.list_schedules()
```
There is no nesting of this sort:
```
versions = mlh.models.versions.list_versions()
```

PR #21, opened 2020-04-07
[CDF-3768] Add FunctionCallsAPI.logs()

HMEiding:
This PR adds the method `FunctionCallsAPI.logs()` which returns logs for a specific function call. It is used as follows:
```
c = CogniteClient()
logs = c.functions.calls.logs(call_id=456, function_id=123)
```
where the logs looks like:
````
print(logs)
[
    {
        "timestamp": "2020-04-07 17:03:08",
        "message": "Started function execution"      
    },
    {
        "timestamp": "2020-04-07 17:03:12",
        "message": "Finished function execution"      
    },
]
````
Optionally, the argument `function_id` can be replaced by `function_external_id`.

The logs can also be retrieved directly from a `FunctionCall` object, like so:
```
func = c.functions.create(name="myfunction", folder="path/to/code")
call = func.call()
logs = call.logs()
```

### DIFF
--- a/cognite/experimental/_api/functions.py
+++ b/cognite/experimental/_api/functions.py
@@ -8,11 +8,6 @@ from cognite.client._api_client import APIClient
 from cognite.experimental.data_classes import Function, FunctionCall, FunctionCallList, FunctionCallLog, FunctionList
 
 
-def _add_function_id_to_calls(calls: FunctionCallList, function_id: int) -> None:
-    for call in calls:
-        call._function_id = function_id
-
-
 class FunctionsAPI(APIClient):
     _RESOURCE_PATH = "/functions"
     _LIST_CLASS = FunctionList
@@ -211,9 +206,7 @@ class FunctionsAPI(APIClient):
         if data:
             body = {"data": data}
         res = self._post(url, json=body)
-        call = FunctionCall._load(res.json(), cognite_client=self._cognite_client)
-        _add_function_id_to_calls([call], id)
-        return call
+        return FunctionCall._load(res.json(), function_id=id, cognite_client=self._cognite_client)
 
     def _zip_and_upload_folder(self, folder, name) -> int:
         current_dir = os.getcwd()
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: cognite/experimental/_api/functions.py [hunk 1]
```diff
@@ -296,4 +305,42 @@ def retrieve(
             function_id = self._cognite_client.functions.retrieve(external_id=function_external_id).id
         url = f"/functions/{function_id}/calls/{call_id}"
         res = self._get(url)
-        return FunctionCall._load(res.json(), cognite_client=self._cognite_client)
+        call = FunctionCall._load(res.json(), cognite_client=self._cognite_client)
+        _add_function_id_to_calls([call], function_id)
```
HMEiding, line 309: Before returning objects of type `FunctionCall` (or `FunctionCallList`), I add the `function_id` as an attribute to this object. This is perhaps i bit hacky, but this information is necessary if we later shall be able to do the following:
```
call = c.functions.calls.retrieve(call_id=456, function_id=123)
logs = call.logs()
```

Should we consider one of the following options to avoid this?
1. Make `function_id` part of the response body from `GET functions/{functionId}/calls` and `GET functions/{functionId}/calls/{callId}`.

or

2. Make calls available without reference to a specific function, like so: `GET functions/calls` and `GET functions/calls/{callId}`.

cognite/experimental/data_classes/functions.py
```diff
@@ -90,9 +91,32 @@ def __init__(
         self.response = response
         self.status = status
         self.error = error
+        self._function_id = function_id
```
HMEiding, line 94: As mentioned in the comment above, this is needed in order to do:
```
call = c.functions.calls.retrieve(call_id=456, function_id=123)
logs = call.logs()
```
The underscore signifies that it will not be printed.

sanderland: this means this class can not be generated, then?

cognite/experimental/_api/functions.py
sanderland: I like returning the id, note that datapoints for time series returns the time series id as well.

HMEiding:
> Another alternative is to define your own _load function which takes a function_id in addition to the object -- this is what a contextualization model/job does now to set its status path.
> will approve and let you make your own choice here, since things are coming together nicely

I agree, that feels like a better solution. I have adopted your solution.

cognite/experimental/data_classes/functions.py
HMEiding: Unfortunately... Or, we can simply add the `_function_id` attribute in the `FunctionCall._load()` and ``FunctionCallList._load()` methods without having it here? I guess that should work as well?

PR #25, opened 2020-04-08
Release version 0.5.0

HMEiding:


andeplane:
Are there any breaking changes here?

andeplane:
Maybe others than functions. Nice!

andeplane:
But I suppose since the older versions aren't going away, it's fine :)

andeplane:
I don't see any value of keeping 0.x, so let's just use regular versioning. major bump = breaking.

andeplane:
0.x is breaking as well! yay

PR #31, opened 2020-04-20
ml entity matcher

sanderland:


cognite/experimental/_api/entity_matching.py
```diff
@@ -18,6 +19,31 @@ def fit(self, items: List[str]) -> EntityMatchingModel:
...
+        true_matches: List[Tuple[int, int]],
+        model_type=None,
+    ) -> EntityMatchingModel:
+        """Fit entity matching model with machine learning methods.
+
+        Args:
+            match_from: entities to match from, should have an 'id' field. Tolerant to passing more than is needed or used (e.g. json dump of time series list)
+            match_to: entities to match to, should have an 'id' field.  Tolerant to passing more than is needed or used.
+            true_matches: Known valid matches given as a list of (id_from,id_to)
+            model_type: model type that defines features and methods used, see API docs for details.
```
andeplane, line 35: Does it work with the default value `None`? If not, we shouldn't have a default value `None`.

cognite/experimental/data_classes/contextualization.py
```diff
@@ -114,6 +114,31 @@ def predict(self, entities: Iterable[str]) -> ContextualizationJob:
...
+            job_path=f"/predictml",
+            match_from=self.dump_entities(match_from),
+            match_to=self.dump_entities(match_to),
+            model_id=self.model_id,
+        )
+
+    @staticmethod
+    def dump_entities(entities: List[Union[Dict, CogniteResource]]):
+        return [
+            {k: v for k, v in e.dump(camel_case=True).items() if isinstance(v, str) or k == "id"}
```
andeplane, line 136: Just to be clear: this will ignore all non-string fields and `id`?

cognite/experimental/_api/entity_matching.py
sanderland: it does

cognite/experimental/data_classes/contextualization.py
sanderland: yes, for now. we can extend this to "flatten" metadata and such later, but for now I did this to limit request size.

PR #32, opened 2020-04-21
[CDF-4863] Add error to data class Function and bump to version 0.5.5

HMEiding:
This is due to the following change in the functions API:
https://github.com/cognitedata/context-api/pull/395


HMEiding:
> errorMessage or error_message?

Neither, it's actually message - the same as for calls. Will fix in the CHANGELOG.md.

PR #38, opened 2020-04-29
assets playground fix

sanderland:
after starbase fixed their error, we don't need this redirection to v1 for retrieve anymore

### DIFF
--- a/cognite/experimental/_api/assets.py
+++ b/cognite/experimental/_api/assets.py
@@ -20,11 +20,14 @@ class ExperimentalAssetsAPI(AssetsAPI):
         chunk_size: int = None,
         name: str = None,
         parent_ids: List[int] = None,
+        parent_external_ids: List[str] = None,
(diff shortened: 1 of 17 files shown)


cognite/experimental/_api/assets.py
```diff
@@ -150,7 +149,6 @@ def retrieve(self, id: Optional[int] = None, external_id: Optional[str] = None)
         utils._auxiliary.assert_exactly_one_of_id_or_external_id(id, external_id)
         return self._retrieve_multiple(ids=id, external_ids=external_id, wrap_ids=True)
 
-    @use_v1_instead_of_playground
     def retrieve_multiple(
```
HMEiding, line 152: It is still needed for `delete`?

sanderland: ah, I don't know, but it should have no effect, while retrieve messes with types/.

PR #39, opened 2020-05-03
Make correct PyPI dependency

pavelzubarev:
`typing-extensions` is a correct dependency name in PyPI


andeplane:
This PR updates more than just that. Is that intentional?

pavelzubarev:
> This PR updates more than just that. Is that intentional?

`poetry lock` did that... I don't know how to make only one thing to be updated

sanderland:
updates are fine. :) thanks for the fix

PR #40, opened 2020-05-04
actions

sanderland:
moving to github actions
* codecov token needs to be added to secrets - @andreavs has it? done


sanderland:
@CsabaAlmasi it's not important for this, using 3.6 now :)

PR #41, opened 2020-05-04
[CDF-3879] FunctionsAPI.create() handles name with / character

HMEiding:
Ideally we would have a test for this, which would be an integration tests. However, we don't have integration tests for functions yet, primarily because creating functions takes some time.


pyproject.toml
```diff
@@ -18,7 +18,7 @@ line_length=120                # corresponds to -w  flag
 multi_line_output=3            # corresponds to -m  flag
 include_trailing_comma=true    # corresponds to -tc flag
 skip_glob = '^((?!py$).)*$'    # this makes sort all Python files
-known_third_party = ["black", "pytest", "requests", "responses", "typing-extensions"]
+known_third_party = ["black", "pytest", "requests", "responses", "typing_extensions"]
```
HMEiding, line 21: Not sure why, but this was forced by the pre-commit hooks.

sanderland: @pavelzubarev changed this the other way -- let's wait for him on why

pavelzubarev: that was changed because typing_extension doesn't exist on PyPi, so my dependency check was failing.

sanderland: but this is an isort parameter, not a dependency

sanderland: yeah confirmed he search-and-replaced a bit too aggressively :D merge ahead @HMEiding

PR #57, opened 2020-05-29
[CDF-6824] Add method to retrieve function response + remove response from FunctionCall

HMEiding:


andeplane:
Can you show in the description how the SDK now should be used?

tests/tests_unit/test_api/test_functions.py
```diff
@@ -478,3 +488,21 @@ def test_get_logs_on_created_call_object(self, mock_function_call_logs_response)
...
+        res = FUNCTION_CALLS_API.get_response(call_id=5678, function_id=FUNCTION_ID)
+        assert isinstance(res, FunctionCallResponse)
+        assert mock_function_call_response_response.calls[0].response.json() == res.dump(camel_case=True)
+
+    @pytest.mark.usefixtures("mock_function_calls_retrieve_response")
+    def test_get_response_on_retrieved_call_object(self, mock_function_call_response_response):
+        call = FUNCTION_CALLS_API.retrieve(call_id=5678, function_id=FUNCTION_ID)
+        response = call.get_response()
+        assert isinstance(response, FunctionCallResponse)
+        assert mock_function_call_response_response.calls[1].response.json() == response.dump(camel_case=True)
```
matzhaugen, line 508: It's not clear to me why we use the second entry here. In general having these magic indices make the test less legible.

HMEiding: I agree. Not sure what the best way to do these mocks is, though.

PR #60, opened 2020-06-02
[CDF-6850] Change return type in functions.retrieve, remove optional

matzhaugen:
The retrieve function method does not return an optional function, only a function. If the requested function fails, the method will raise an error.

### DIFF
--- a/cognite/experimental/_api/functions.py
+++ b/cognite/experimental/_api/functions.py
@@ -145,7 +145,7 @@ class FunctionsAPI(APIClient):
             external_id (str, optional): External ID
 
         Returns:
-            Function: Requested function or None if it does not exist.
+            Function: Requested function.
         Raises:
             ~cognite.client.exceptions.CogniteAPIError: Function id not found
 
--- a/cognite/experimental/_api/functions.py
+++ b/cognite/experimental/_api/functions.py
@@ -137,7 +137,7 @@ class FunctionsAPI(APIClient):
         res = self._get(url)
         return FunctionList._load(res.json()["items"], cognite_client=self._cognite_client)
 
-    def retrieve(self, id: Optional[int] = None, external_id: Optional[str] = None) -> Optional[Function]:
+    def retrieve(self, id: Optional[int] = None, external_id: Optional[str] = None) -> Function:
         """`Retrieve a single function by id. <https://docs.cognite.com/api/playground/#operation/get-api-playground-projects-project-functions-function_name>`_
 
         Args:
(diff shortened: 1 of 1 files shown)


HMEiding:
Isn't it better to perhaps return `None` and keep the `Optional`? At least that seems to be how it's done for assets.

matzhaugen:
The cleanest way would be to have an extra arg like they do here: https://github.com/cognitedata/cognite-sdk-python/blob/8a8cc104373e47b74bcfe0a0582c61afa8e698b1/cognite/client/_api/assets.py#L370

Otherwise, I could wrap it into a try/except clause. What do you think?

HMEiding:
I looked into this a while back, and it's coming back to me now. We're using the `_retrieve_multiple()` method which we inherit from the `APIClient`, which calls the `/byids` endpoint:
https://docs.cognite.com/api/playground/#operation/post-api-playground-projects-project-context-functions-byids

The `retrieve_multiple()` function looks at the response code from `/byids`, and if it is a 400 AND it gives a list of missing IDs, then it throws a `CogniteNotFoundError`. This is done inside this function:
https://github.com/cognitedata/cognite-sdk-python/blob/8a8cc104373e47b74bcfe0a0582c61afa8e698b1/cognite/client/_api_client.py#L262
The `CogniteNotFoundError` is handled by the try-except, as you can see, and `None` is returned.
However, if the response code is a 400 and there is no list of missing IDs, it throws a `CogniteAPIError`, and this is not handled on the outside. Thus, I think the proper handling is to add the list of missing IDs in the context-api. This is what for example assets does, as you can see here:
https://docs.cognite.com/api/playground/#operation/byIdsAssets

HMEiding:
After the fix in the context-api, what's the status of this now @matzhaugen ?

matzhaugen:
I'm not sure exactly what we want the behavior here to be. Should we return an error from the API, and if so it will already have the missing field. Otherwise we can return our own error in the sdk with it's own missing field, but I'm not sure if there is a precedent for that.

HMEiding:
> I'm not sure exactly what we want the behavior here to be. Should we return an error from the API, and if so it will already have the missing field. Otherwise we can return our own error in the sdk with it's own missing field, but I'm not sure if there is a precedent for that.

HMEiding:
I think we should be consistent with the rest of the SDK, like for example assets, which has the `Optional`. With the recent PR on the context-api, I believe that this should be the current behaviour (whereas before the PR, we got an exception). You could test and see?

matzhaugen:
Yes, I just checked and the behavior is the same assets now. We can close this PR

PR #65, opened 2020-06-19
[CDF-4730] Add update() method on Function class

HMEiding:
Now we can do:
```
function = c.functions.create(name=”func”, folder=”folder”)
while function.status in ["Queued", "Deploying"]:
      function.update()
      sleep(1)
```

### DIFF
--- a/cognite/experimental/data_classes/functions.py
+++ b/cognite/experimental/data_classes/functions.py
@@ -106,7 +106,7 @@ class Function(CogniteResource):
         return self._cognite_client.functions.calls.retrieve(call_id=id, function_id=self.id)
 
     def update(self) -> None:
-        """Update the funciton object.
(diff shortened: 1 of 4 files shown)


cognite/experimental/data_classes/functions.py
```diff
@@ -105,6 +105,22 @@ def retrieve_call(self, id: int) -> "FunctionCall":
         """
         return self._cognite_client.functions.calls.retrieve(call_id=id, function_id=self.id)
 
+    def update(self) -> None:
+        """Update the function object.
```
matzhaugen, line 109: Why/when would you update the function object? When adding attributes to it? Maybe elaborate a bit in the comments.

HMEiding: Agree

PR #74, opened 2020-08-05
Add function_path argument to create function

bjornarhem:
This PR add a function_path as an argument to `client.functions.create`. Now you can:

```python
client = CogniteClient(...)

function = c.functions.create(name="myfunction", folder="path/to/code", function_path="path/to/function.py")

```

`function_path` is relative to the folder you specify.


tests/tests_unit/test_api/test_functions.py
```diff
@@ -197,20 +198,28 @@ def mock_function_calls_list_response(rsps):
 
 class TestFunctionsAPI:
     @pytest.mark.parametrize(
-        "function_folder, will_pass",
-        [("function_code", True), ("bad_function_code", False), ("bad_function_code2", False)],
+        "function_folder, function_path, exception",
+        [
+            ("function_code", "./handler.py", None),
```
matzhaugen, line 203: Can we also have a test case for `function_folder="."`? This is the most common case, I presume

PR #75, opened 2020-08-05
detect for both doc and p&id parsing, min_tokens added as a parameter

olacognite:
CDF-8292


sanderland:
Look good. Ping me when you get tests corrected + version bump + changelog

PR #79, opened 2020-08-10
CDF-8165, Add model info entity matching

rebwib:
This PR updates retrieve so that it also outputs the model info for entity matcher models.
It solves https://cognitedata.atlassian.net/browse/CDF-8165

### DIFF
--- a/cognite/experimental/data_classes/contextualization.py
+++ b/cognite/experimental/data_classes/contextualization.py
@@ -91,6 +91,10 @@ class ContextualizationModel(CogniteResource):
         start_timestamp=None,
         status_timestamp=None,
         cognite_client=None,
+        classifier=None,
+        feature_type=None,
+        keys_from_to=None,
+        model_type=None,
     ):
         self.model_id = model_id
         self.status = status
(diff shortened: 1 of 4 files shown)


cognite/experimental/data_classes/contextualization.py
```diff
@@ -91,13 +91,21 @@ def __init__(
...
         self.model_id = model_id
         self.status = status
         self.request_timestamp = request_timestamp
         self.start_timestamp = start_timestamp
         self.status_timestamp = status_timestamp
         self.error_message = error_message
+        self.classifier = classifier
+        self.feature_type = feature_type
+        self.keys_from_to = keys_from_to
+        self.model_type = model_type
```
sindrest-cognite, line 108: The `ContextualizationModel` class is pretty generic (it could cover e.g a future classification model or object detection model), but the attributes you are adding here are entity matching specific. I'm wondering if we instead should add this model specific data as a more general `metadata` field or similar.

rebwib: I agree, see updated changes.

PR #80, opened 2020-08-13
[CDF-8432], [CDF-8461] Update list-method

rebwib:
This PR solves https://cognitedata.atlassian.net/browse/CDF-8432?atlOrigin=eyJpIjoiMGQ0MzJhYzQ0YzQwNDc0YjkwOGY5YTNjZGQ1YjYzOTciLCJwIjoiaiJ9
and https://cognitedata.atlassian.net/browse/CDF-8461?atlOrigin=eyJpIjoiYjdjM2FjNWVlMTFmNDNkMTlkOTZkNThkZmI2MGJhOTciLCJwIjoiaiJ9
It changes the list method so that it only lists models and adds the possibility to filter the listed models based on parameter values.


cognite/experimental/_context_client.py
```diff
@@ -67,14 +67,26 @@ def list(self) -> ContextualizationModelList:
...
-            ContextualizationModelList: List of models."""
+            ContextualizationModelList: List of jobs."""
         return ContextualizationModelList(
             [
                 self._MODEL_CLASS._load(model, cognite_client=self._cognite_client)
                 for model in self._camel_get("/").json()["items"]
             ]
         )
 
+    def list_models(self) -> ContextualizationModelList:
```
sindrest-cognite, line 78: I'm thinking we should instead change the existing `list` method to list only models, and maybe add a new method `list_jobs` to list all jobs. Thoughts?

rebwib: I thought about the same thing. Did it like this to not change the behaviour of the current `list`. 
Maybe remove `list` and have `list_jobs` and `list_models`? 

Let's here what @sanderland thinks when he is back next week.

sanderland: I agree with sindre, list should list what retrieve returns, etc. We can bump to 0.16 to break.

rebwib: Sounds good, see my changes in 6f203e0.

cognite/experimental/_context_client.py
```diff
@@ -63,15 +63,32 @@ def retrieve(self, model_id: int) -> ContextualizationModel:
...
-    def list(self) -> ContextualizationModelList:
+    def list(self, filter: Dict = None) -> ContextualizationModelList:
         """List models
 
+        Args:
+            filter (dict): If not None, return models with parameter values that matches what is specified in the filter.
+
         Returns:
             ContextualizationModelList: List of models."""
+        if filter:
```
sanderland, line 74: just post is fine, consistent with other sdk stuff which basically ignores /

PR #81, opened 2020-08-13
Bump SDK to 0.15.2: Insert a sleep cycle for files api queries in functions creation

matzhaugen:


cognite/experimental/_api/functions.py
```diff
@@ -100,6 +101,20 @@ def create(
             _validate_function_handle(function_handle)
             file_id = self._zip_and_upload_handle(function_handle, name)
 
+        sleep_time = 1.0 # seconds
+        for i in range(MAX_RETRIES):
+            try:
+                self._cognite_client.files.retrieve(id=file_id)
```
HMEiding, line 107: 1. `self._cognite_client.files.retrieve(id=file_id)` returns `None` if the file does not exist. It does not raise.
2. Although checking file existence is useful as well, what we actually need is to check that the file contents is uploaded; the file can exist (i.e. the metadata), even though the file contents does not.

PR #86, opened 2020-08-19
[CDF-8458] Entity matching update models, retrieve (byids)

rebwib:
This solves https://cognitedata.atlassian.net/browse/CDF-8458?atlOrigin=eyJpIjoiYzFmNjA0ZjYyYmU2NDAxYWFjMjM4NGZlMmRkODdmZjAiLCJwIjoiaiJ9

@sanderland I was not sure what to return. Hope you can give me feedback on whether we should return a list or a model object.


cognite/experimental/_context_client.py
```diff
@@ -53,7 +53,7 @@ def _fit_model(self, model_path="/fit", headers=None, **kwargs) -> Union[EntityM
         response = self._camel_post(model_path, json=kwargs, headers=headers)
         return self._MODEL_CLASS._load(response.json(), cognite_client=self._cognite_client)
 
-    def retrieve(self, model_id: int) -> ContextualizationModel:
+    def retrieve(self, model_id: int) -> Union[EntityMatchingModel]:
```
sanderland, line 56: union with what?

sanderland (changes requested):
the update call itself should return a list of updated objects, so you should not need to have to call retrieve either before or after. this should only be a single http call.

the sdk takes List[Union[Asset, AssetUpdate]] which is quite involved with autogeneration magic etc.
I would suggest taking a list of models and just throw the entire model dumps at the update endpoint, and have the update endpoint complain if some fields are set that both change and aren't updateable.   is there something in the api design for this?

cognite/experimental/_context_client.py
rebwib: After we removed resource typing it is alway  `EntityMatchingModel` so I can remove the union.

rebwib:
> the update call itself should return a list of updated objects, so you should not need to have to call retrieve either before or after. this should only be a single http call.
> 
> the sdk takes List[Union[Asset, AssetUpdate]] which is quite involved with autogeneration magic etc.
> I would suggest taking a list of models and just throw the entire model dumps at the update endpoint, and have the update endpoint complain if some fields are set that both change and aren't updateable. is there something in the api design for this?

I tried to follow how the SDK updates for instance an asset when i created this method. However, was a bit unsure about the response. Changed it in my last commit. As of now, it is only possible to update one model with each update-call. The reason was to resemble update in the SDK.

There is a section on update in the API design, https://github.com/cognitedata/service-contracts/blob/master/design_guide/API_design_guide.md#update I tried to follow that when I created the update endpoint.

cognite/experimental/_context_client.py
```diff
@@ -89,6 +89,27 @@ def list_jobs(self) -> ContextualizationModelList:
...
+            key
+            for key in model_attributes.keys()
+            if model_attributes[key] != model_update_attributes[key]
+            and key not in ContextualizationModel().__dict__.keys()
+        ]
+        update_dict = {}
+        for attribute in attributes_update:
+            update_dict[attribute] = {"set": model_update_attributes[attribute]}
+
+        return self._camel_post("/update", json={"items": [{"modelId": model.model_id, "update": update_dict}]}).json()
```
sanderland, line 111: definitely convert this to an EntityMatchingModel again though :)

sanderland:
> > the update call itself should return a list of updated objects, so you should not need to have to call retrieve either before or after. this should only be a single http call.
> > the sdk takes List[Union[Asset, AssetUpdate]] which is quite involved with autogeneration magic etc.
> > I would suggest taking a list of models and just throw the entire model dumps at the update endpoint, and have the update endpoint complain if some fields are set that both change and aren't updateable. is there something in the api design for this?
> 
> I tried to follow how the SDK updates for instance an asset when i created this method. However, was a bit unsure about the response. Changed it in my last commit. As of now, it is only possible to update one model with each update-call. The reason was to resemble update in the SDK.
> 
> There is a section on update in the API design, https://github.com/cognitedata/service-contracts/blob/master/design_guide/API_design_guide.md#update I tried to follow that when I created the update endpoint.

Don't really like the retrieve call, but alright then, efficiency shouldn't be that important here.

cognite/experimental/_context_client.py
rebwib: See 5d8f473. If that was not correct, it would be great if you could push a fix to this branch.

rebwib:
> > > the update call itself should return a list of updated objects, so you should not need to have to call retrieve either before or after. this should only be a single http call.
> > > the sdk takes List[Union[Asset, AssetUpdate]] which is quite involved with autogeneration magic etc.
> > > I would suggest taking a list of models and just throw the entire model dumps at the update endpoint, and have the update endpoint complain if some fields are set that both change and aren't updateable. is there something in the api design for this?
> > 
> > 
> > I tried to follow how the SDK updates for instance an asset when i created this method. However, was a bit unsure about the response. Changed it in my last commit. As of now, it is only possible to update one model with each update-call. The reason was to resemble update in the SDK.
> > There is a section on update in the API design, https://github.com/cognitedata/service-contracts/blob/master/design_guide/API_design_guide.md#update I tried to follow that when I created the update endpoint.
> 
> Don't really like the retrieve call, but alright then, efficiency shouldn't be that important here.

I didn't really like it either, but wasn't sure how else to do it.

sanderland:
in the sdk it's
```
    def _get_update_properties(cls):
        return [key for key in cls.__dict__.keys() if (not key.startswith("_")) and (not key == "labels")]
```
i.e. everything is updateable and the backend ignores things like id if they're the same

rebwib:
> ef _get_update_properties(cls):
>         return [key for key in cls.__dict__.keys() if (not key.startswith("_")) and (not key == "labels")]

Then we have to change the api as well, since for now it only lets you update name and description. Should we let the user update e.g. feature_type?

sanderland (approved):

PR #91, opened 2020-09-01
entitymatching predict/refit without having an object already

sanderland:


cognite/experimental/_api/entity_matching.py
```diff
@@ -139,6 +139,51 @@ def fit(
...
+            complete_missing=complete_missing,
+        )
+
+    def refit(
+        self, true_matches: List[Tuple[int, int]], id: Optional[int] = None, external_id: Optional[str] = None
+    ) -> "EntityMatchingModel":
+        """Re-fits an entity matching on updated data.
+
+        Args:
+            true_matches: Updated known valid matches given as a list of (id_from,id_to).
```
rebwib, line 180: Should we instead call this additional known valid matches given, since the true matches from fit and refit are merged?

rebwib (approved):
I think it should be made clear that refit combines the true matches. Other than that, looks good to me.

PR #100, opened 2020-09-29
Relationship: Add high-level query feature for Relationship

Tapped:
Add a high-level relationship query resolver, to make it easier to traverse relationships.
It also supports filtering by label on the resolved resources, to allow saying match all relationships where "Asset has label Pump" for example.

### DIFF
--- a/cognite/experimental/__init__.py
+++ b/cognite/experimental/__init__.py
@@ -1,1 +1,2 @@
 from cognite.experimental._client import CogniteClient
+from cognite.experimental._relationships_query import RelationshipsQuery
--- a/cognite/experimental/_api/relationships.py
+++ b/cognite/experimental/_api/relationships.py
@@ -3,6 +3,13 @@ from typing import Any, Dict, Generator, List, Optional, Set, Union
 
 from cognite.client import utils
 from cognite.client._api_client import APIClient
+from cognite.client.data_classes.labels import LabelFilter
+from cognite.experimental._relationships_query import (
+    RelationshipsQuery,
+    RelationshipWithResource,
+    Resource,
+    ResourceFilter,
+)
 from cognite.experimental.data_classes import Relationship, RelationshipFilter, RelationshipList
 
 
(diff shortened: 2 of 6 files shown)


cognite/experimental/_relationships_query.py
```diff
@@ -0,0 +1,181 @@
...
+        self.retrieve_multiple = retrieve_multiple
+
+    def retrieve_multiple(self, external_ids: List[str]):
+        if self.has_ignore_unknown_ids:
+            return self.retrieve_multiple(external_ids=external_ids, ignore_unknown_ids=True)
+        else:
+            return self.retrieve_multiple(external_ids=external_ids)
+
+
+class RelationshipsQuery:
```
Tapped, line 56: I'm very open to suggestion for how to expose this now. I don't like this `RelationshipsQuery(c).query` thingy. We might even just expose the function itself, and let it take in the client as argument.

Tapped: What do you think @sanderland ?

sanderland: c.relationships.query?

Tapped: I would love if the beta client was relying on composition instead though, since then this would be easy.

Tapped: We might add this feature to the main package instead ofc.

sanderland: isn't relationships beta anyway, and experimental about to be removed from the main sdk altogether?

Tapped: Idk 🤷‍♂️ this is more like an experimental SDK feature, and not tied to the stability of relationships.

Tapped: However, I'll likely move it tomorrow, so that it's easier to integrate it.

Tapped:
@sanderland if we just ignore how we expose this feature, could you review the rest of the code and see if it looks sane?

I'm thinking of adding a bit more tests as well (Especially if this is going into main SDK).

sanderland:
> @sanderland if we just ignore how we expose this feature, could you review the rest of the code and see if it looks sane?
> 
> I'm thinking of adding a bit more tests as well (Especially if this is going into main SDK).

overall it doesn't look very pythonesque, more like someone used to different languages converted it into python. it's hard to review

Tapped:
But specifically, what makes it not pythonesque?
* I'm relying on inheritance (ugh, but that's pytonesque right?)
* I do snake_casing for all the variable names.
* I'm using list interpolator when it makes sense.
* I'm using the ugly typing system of Python so that I'm able to get some decent code-suggestions
* I'm using python documentation syntax
* I let a method have 5+ arguments that are by default set to null, instead of having a class for my arguments
* I'm using if expressions

Please teach me, instead of just stating it's not pythonesque.

cognite/experimental/_relationships_query.py
```diff
@@ -0,0 +1,187 @@
...
+        else:
+            return False
+
+
+class RelationshipFilterAll(RelationshipFilterBase):
+    def is_inside(self, resource: Resource):
+        return True
+
+
+class RelationshipWithData(CogniteResource):
```
Tapped, line 37: I guess I'll rename this to `RelationshipWithResource`

cognite/experimental/_relationships_query.py
erlendvollset: Does cognite.client.beta.CogniteClient contain all functionality from cognite.client.CogniteClient? If so, could you just have `cognite.experimental._client.CognieClient` inherit from the beta client instead and add this directly to `experminental.client.relationships?`

Tapped: The cognite.client.beta.CogniteClient derives from cognite.client.CogniteClient, so we might go for that approach. However, it's a bit icky that the `experimental.client.relationships` is today using playground, while this feature above uses `beta`. However, it's only going to be like this for a while I guess.

Tapped: So now I moved it behind `experimental.client.relationships.query`

cognite/experimental/__init__.py
```diff
@@ -1 +1,2 @@
 from cognite.experimental._client import CogniteClient
+from cognite.experimental._relationships_query import RelationshipsQuery
```
erlendvollset, line 2: Don't need this anymore.

cognite/experimental/_relationships_query.py
```diff
@@ -0,0 +1,128 @@
...
+    def is_inside(self, resource: Resource):
+        return True
+
+
+class RelationshipWithResource(CogniteResource):
+    def __init__(self, relationship: Relationship, source_resource: Resource, target_resource: Resource):
+        self.relationship = relationship
+        self.source_resource = source_resource
+        self.target_resource = target_resource
+
```
erlendvollset, line 42: Move this into cognite.expermiental.data_classes.experimental namespace?

cognite/experimental/_api/relationships.py
```diff
@@ -351,3 +358,81 @@ def delete(self, external_id: Union[str, List[str]]) -> None:
...
+
+                Use result of a query::
+
+                    >>> from cognite.client.beta import CogniteClient
+                    >>> from cognite.experimental import RelationshipsQuery
+                    >>> c = CogniteClient()
+                    >>> rel_query = RelationshipsQuery(c)
+                    >>> result = rel_query.query()
+                    >>> for res in result:
+                    >>>     print(res.source_resource.name, res.target_resource.name)
```
erlendvollset, line 426: update examples to do c.relationships.query() instead

Tapped: Great catch 👍

sanderland:
Doesn't seem this fits in the experimental sdk after all, which is primarily for alpha/playground stuff and relationships are being removed (see the move to .relationships_playground). Suggest we close this here.
Perhaps the main sdk is better, but after checking with @andreavs and the power-sdk, the need doesn't seem clear, so I'm not sure it fits there either. maybe check with the relationships team first.

Tapped:
Yeah, makes sense. I think I'll separate the logic of listing relationships from this function. The property that creates the most value is the automatic resolving of resources by relations. This is what there will most likely be more demand for now that's relationship will start to receive more usage. This function is very similar to then the `asset.timeseries()` or `asset.events()` functions we provide.

PR #106, opened 2020-10-13
Add data_set_id to Functions.create

haakonvt:
Let me know if this is okey, it would make it much easier to work with for e.g. AIR 😄  @senickel

### DIFF
--- a/cognite/experimental/_api/functions.py
+++ b/cognite/experimental/_api/functions.py
@@ -1,5 +1,4 @@
 import importlib.util
-import json
 import os
 import sys
 import time
@@ -7,7 +6,7 @@ from inspect import getsource
 from numbers import Number
 from pathlib import Path
 from tempfile import TemporaryDirectory
-from typing import Any, Callable, Dict, List, Optional, Union
+from typing import Callable, Dict, List, Optional, Union
 from zipfile import ZipFile
 
 from cognite.client import utils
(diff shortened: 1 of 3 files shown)


cognite/experimental/_api/functions.py
```diff
@@ -49,6 +48,7 @@ def create(
         secrets: Optional[Dict] = None,
         cpu: Number = 0.25,
         memory: Number = 1.0,
+        data_set_id: Optional[int] = None,
```
pavelzubarev, line 51: The big question for me should we ever expose internal IDs vs. external IDs. I believe external IDs are the way to go and we should not require client code to resolve that explicitly. What do you think, @haakonvt

haakonvt: Boy if I agree ;) But have a look at the FilesAPI.upload; it only accepts `data_set_id` 😢

senickel: or do both?

HMEiding (changes requested):
I think adding the argument `data_set_id` to `FunctionsAPI.create()` is a bit misleading since it suggests that the `data_set_id` is referring to the _**function**_ and not the file. I would much rather have for example a helper function `FunctionsAPI.upload_file_from_folder()` which could have `data_set_id` as an optional argument and return a `file_id`.

haakonvt:
> I think adding the argument `data_set_id` to `FunctionsAPI.create()` is a bit misleading since it suggests that the `data_set_id` is referring to the _**function**_ and not the file. I would much rather have for example a helper function `FunctionsAPI.upload_file_from_folder()` which could have `data_set_id` as an optional argument and return a `file_id`.

I agree and I like it! I'll see what I can make 🚀

haakonvt:
I'll abandon this one, since branch name is `add-data-set-id-to-functions-create`

PR #109, opened 2020-10-20
plot data extraction

sanderland:


minshi-cognite (approved):
You updated entity matching test by the way in this pr :)

PR #112, opened 2020-11-09
Em pipelines fixes

sanderland:


chingyu-cognite (approved):
Looks good. 

The default value of the` limit` is different from the docstring in both `list` methods.
https://github.com/cognitedata/cognite-sdk-python-experimental/blob/empipelinesupdate/cognite/experimental/_api/entity_matching.py#L40

PR #113, opened 2020-11-12
[CDF-9977] Raise if unauthorised

rebwib:
Fix to not exposed stack trace like when error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@ Changes are grouped as follows
 - `Security` in case of vulnerabilities.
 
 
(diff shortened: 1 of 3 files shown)


sanderland:
Why turn a CogniteAPIError in to a plain Exception? It seems like whatever uses the SDK should handle this instead.

rebwib:
> Why turn a CogniteAPIError in to a plain Exception? It seems like whatever uses the SDK should handle this instead.

Currently if a user does not have the correct access right the entire stack trace is returned, https://cognitedata.atlassian.net/browse/CDF-9977 

Do you have a better suggested solution?

PR #123, opened 2020-12-07
[CDF-10142] Add integrations and runs APIs

AlexThunder:


INTEGRATIONS&RUNS.MD
```diff
@@ -0,0 +1,40 @@
+# Examples for the Integrations and Runs APIs
+Create a client and try next possible operations
```
wjoel, line 2: Just leave it at "Create a client" as in `CONTEXTUALIZATION.md`. Add a note explaining that the API key must have read/write permissions for the "generic" ACL.

INTEGRATIONS&RUNS.MD
```diff
@@ -0,0 +1,40 @@
...
+```python
+from cognite.experimental import CogniteClient
+
+client = CogniteClient(api_key="<<itera-int-green api key>>", project="itera-int-green", base_url = "https://greenfield.cognitedata.com")
```
wjoel, line 6: Shouldn't refer to itera-int-green, just "<<your-api-key>>" and "your-project". Remove base_url.

INTEGRATIONS&RUNS.MD
```diff
@@ -0,0 +1,40 @@
...
+
+# Create integration
```
wjoel, line 8: This should be done outside of the python code block to get nice headers as in https://raw.githubusercontent.com/cognitedata/cognite-sdk-python-experimental/master/CONTEXTUALIZATION.md

INTEGRATIONS&RUNS.MD
```diff
@@ -0,0 +1,40 @@
...
+
+from cognite.experimental.data_classes import Integration	
+i1 = Integration(external_id = "py test id", name="py test", description = "python generated", data_set_id = 1, schedule = "", owner = {"name": "Alex", "email": "Alex@test.no"})
+res = client.integrations.create(i1)
+
+#Get list
```
wjoel, line 14: "List integrations"

INTEGRATIONS&RUNS.MD
wjoel, line 8: "Create an integration"

INTEGRATIONS&RUNS.MD
```diff
@@ -0,0 +1,40 @@
...
+integrations_list = client.integrations.list()
+
+#Update existed integration
```
wjoel, line 17: "Update an integration that already exists"

INTEGRATIONS&RUNS.MD
```diff
@@ -0,0 +1,40 @@
...
+res.description = "New description"
+res = client.integrations.update(res)
+
+#or
+from cognite.experimental.data_classes import IntegrationUpdate	
+up = IntegrationUpdate(id=res.id)
+up.description.set("Another new integration")
+res = client.integrations.update(up)
+
+#getById
```
wjoel, line 27: "Retrieve an integration by id"

INTEGRATIONS&RUNS.MD
```diff
@@ -0,0 +1,40 @@
...
+res1 = client.integrations.retrieve(id=res.id)
+
+#list of runs
```
wjoel, line 30: "List integration runs"

INTEGRATIONS&RUNS.MD
```diff
@@ -0,0 +1,40 @@
...
+up.description.set("Another new integration")
+res = client.integrations.update(up)
+
+#getById
+res1 = client.integrations.retrieve(id=res.id)
+
+#list of runs 
+runsList = client.runs.list(external_id=res1.external_id)
+
+#add new run with status
```
wjoel, line 33: "Add a new run with status" or "Create a new integration run with status"

INTEGRATIONS&RUNS.MD
```diff
@@ -0,0 +1,40 @@
...
+from cognite.experimental.data_classes import IntegrationRun	
+newRun = client.runs.create(IntegrationRun(external_id=res1.external_id, status="success"))
+
+#delete integration with all runs
```
wjoel, line 37: "Delete an integration and all its runs"

cognite/experimental/_api/integrationruns.py
```diff
@@ -0,0 +1,63 @@
...
+                >>> c = CogniteClient()
+                >>> runsList = c.runs.list(external_id="test ext id", limit=5)
+        """
+
+        self._LIST_CLASS = IntegrationWithStatusesList
+        response_list: IntegrationWithStatusesList
+        response_list = self._list(method="GET", limit=limit, filter={"externalId": external_id})
+        if len(response_list) == 1:
+            return response_list[0].runs()
+        return IntegrationRunList([])
```
wjoel, line 37: Should this return the full list? I don't understand this one.

tests/tests_unit/test_api/test_integrations.py
```diff
@@ -0,0 +1,170 @@
...
+                "description": "description",
+                "createdTime": 1565965333132,
+                "lastUpdatedTime": 1565965333132,
+                "dataSetId": 1,
+                "owner": {
+                    "name": "test name",
+                    "email": "aaa@cognite.com"
+                },
+                "metadata": {
+                    "step": 22,
```
wjoel, line 30: Hm, I guess we allow this in the Python SDK even though metadata is a string->string map? Python SDK converts the value to string for the user?

But, for a mock, shouldn't it be string->string as that's what we return from the API?

cognite/experimental/_api/integrationruns.py
AlexThunder: Yes, it should be full list here.
Response is a list of objects with fields:
           external_id: str = None,
            statuses: List[Dict[str, Any]] = None

Where statuses contains list of Runs. So runs() method - returns those list adding externalId inside each of element.

tests/tests_unit/test_api/test_integrations.py
AlexThunder: As we store metadata to JSONB, I have set metadata in python as metadata: Dict[str, Any].
So now if we have integration response part : 
`    "metadata": {
        "aaa": "bbb",
        "ccc": 111
    }`
it is converted to dict in python 
`{'aaa': 'bbb', 'ccc': 111}`

And if we create new integration with this dict. It will be stored to database as JSON : 
`  "metadata": {
        "aaa": "bbb",
        "ccc": 111
    }`

wjoel: That's a bug that we need to fix then, metadata should not allow arbitrary values, but only strings. In addition we have some restrictions on how large the metadata can be that we should also enforce at some point (as soon as possible). Here's the documentation from assets, for example:
`Custom, application specific metadata. String key -> String value. Limits: Maximum length of key is 128 bytes, value 10240 bytes, up to 256 key-value pairs, of total size at most 10240.`

cognite/experimental/_client.py
```diff
@@ -107,3 +109,5 @@ def __init__(
         self.plot_extraction = PlotDataExtractionAPI(self._config, api_version="playground", cognite_client=self)
 
         self.functions = FunctionsAPI(self.config, api_version="playground", cognite_client=self)
+        self.integrations = IntegrationsAPI(self._config, api_version="playground", cognite_client=self)
+        self.runs = IntegrationsRunsAPI(self._config, api_version="playground", cognite_client=self)
```
wjoel, line 113: This should be `integration_runs`

AlexThunder: Done

tests/tests_unit/test_api/test_integrations.py
AlexThunder: Fixed

mathialo (changes requested):
Please bump the (minor) [version number](https://github.com/cognitedata/cognite-sdk-python-experimental/blob/fbabbd0f2a77636cc229bf0066a3cfa757cbf1d8/pyproject.toml#L3), so that a new release is triggered. 

Other than that it seems good to me, great work!

mathialo (approved):

PR #132, opened 2021-01-11
Domain completion

sanderland:


cognite/experimental/_api/schema_completion.py
```diff
@@ -6,14 +6,35 @@
...
-        return self._run_job(job_path="/complete", status_path="/", external_id=external_id,)
+        return self._run_job(job_path="/type", status_path="/", external_id=external_id)
+
+    def complete_domain(
+        self, external_id: str, template_name: str, asset_property: str, version: int = None
+    ) -> ContextualizationJob:
+        """Completes a schema uploaded in CDF as a domain.
+
+        Args:
+            external_id (str), version (int): External ID of the domain to work on, with optional version.
```
rebwib, line 27: How will this look in the documentation. Is it better to have a new line, e.g.:
`external_id (str): External ID of the domain to work on.`
`version (int):  Optional version of the domain.`
?

cognite/experimental/_api/schema_completion.py
```diff
@@ -6,14 +6,35 @@
...
+            template_name: Name of the template to be completed within the domain
```
rebwib, line 28:
```suggestion
            template_name (str): Name of the template to be completed within the domain
```

cognite/experimental/_api/schema_completion.py
```diff
@@ -6,14 +6,35 @@
...
+            asset_property: Which field (with constant type) in the template defines the externalId of the parent asset in each entry.
```
rebwib, line 29:
```suggestion
            asset_property (str): Which field (with constant type) in the template defines the externalId of the parent asset in each entry.
```

rebwib (approved):
Just a couple of suggestions, but looks good to me.

PR #138, opened 2021-01-26
Cdf 10263 match rules

olacognite:


cognite/experimental/data_classes/contextualization.py
```diff
@@ -321,6 +329,43 @@ class EntityMatchingPipelineList(CogniteResourceList):
...
+        return dataframe_summarize(df)._repr_html_()
+
+
+class MatchRulesApplyJob(ContextualizationJob):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self._status_path = "/context/matchrules/suggest/"
+
+    @property
+    def rules(self) -> EntityMatchingMatchRuleList:
```
sanderland, line 355: apply has a rules method and not a matches method?

olacognite: The matches are organized per rule, could maybe do various. rule_information and matches?

sanderland: ah yes, it's fine like this then

sanderland (approved):

PR #144, opened 2021-02-03
[MLOPS-636] Add limit argument to Function.list_calls()

HMEiding:
This PR adds the `limit` argument to `Function.list_calls()`.


cognite/experimental/_constants.py
```diff
@@ -0,0 +1,5 @@
+# Functions
```
HMEiding, line 1: I moved these constants to a separate file in order to avoid circular imports.

matzhaugen (approved):
💯

PR #149, opened 2021-02-11
Add text description pnid parsing

rebwib:
From user interview we got feedback that it should be explicitly stated in the documentation that results are not written to CDF.


olacognite (approved):
Looks better safe than sorry to me

PR #152, opened 2021-02-24
[CDF-10876] Change runs list method

AlexThunder:
Add add-remove functionality for list elements


mathialo (approved):
Very nice! Much cleaner now that the API is more similar to the other endpoints 🙂

PR #154, opened 2021-02-25
Change IntegrationRuns list method

mathialo:
API has been updated to follow the convention of other endpoints (ie, JSON is on the form `{ "items" : [ ... ] }` with pagination). Update SDK accordingly. 

Originally from an external PR (#152), but because of limited access to api-keys in the tests it has to go through an extra branch/PR.

### DIFF
--- a/cognite/experimental/data_classes/integrations.py
+++ b/cognite/experimental/data_classes/integrations.py
@@ -80,6 +80,26 @@ class IntegrationUpdate(CogniteUpdate):
         def set(self, value: Any) -> "IntegrationUpdate":
             return self._set(value)
 
+    class _ObjectIntegrationUpdate(CogniteObjectUpdate):
+        def set(self, value: Dict) -> "IntegrationUpdate":
+            return self._set(value)
+
+        def add(self, value: Dict) -> "IntegrationUpdate":
+            return self._add(value)
+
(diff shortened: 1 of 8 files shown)


cognite/experimental/data_classes/integrations.py
```diff
@@ -80,6 +80,26 @@ class _PrimitiveIntegrationUpdate(CognitePrimitiveUpdate):
         def set(self, value: Any) -> "IntegrationUpdate":
             return self._set(value)
 
+    class _ObjectIntegrationUpdate(CogniteObjectUpdate):
+        def set(self, value: Dict) -> "IntegrationUpdate":
```
emelvarol-cognite, line 84: I think I don't know this syntax. I thought -> shows the return type?

tests/tests_unit/test_api/test_runs.py
```diff
@@ -18,17 +13,10 @@
 def mock_run_response(rsps):
```
emelvarol-cognite, line 13: Where do you use this?

emelvarol-cognite (approved):
Looks good

cognite/experimental/data_classes/integrations.py
sanderland: quoting a type hint is done for classes/types that are not yet defined.

PR #155, opened 2021-02-25
[MLOPS-679] Setting `limit=None` in `Function.list_schedules`

qTipTip:
The `function.list_schedules` method was not updated accordingly when we changed the `limit`-defaults. We now pass `limit=None` to make sure we always get every schedule associated to the relevant function.﻿


pavelzubarev (approved):
LGTM

PR #160, opened 2021-03-18
Fix missing client from _load in func.list and schedules.list

haakonvt:
As reported by @senickel on Slack: https://cognitedata.slack.com/archives/C010VCGKBCM/p1616056126005500

Bumping version to `0.48.2`


HMEiding (approved):
LGTM

PR #180, opened 2021-06-22
Cdf 12366 update diagram

chingyulin:
https://github.com/cognitedata/cognite-sdk-python/pull/830
from stable sdk: in a recent update in context-api, we introduced three more statues (distributed, distributing and collecting), which breaks the current setup and test in experimental sdk, this pr introduces these statuses and refactor them as constants


olacognite (commented):
Would be better to have one contextualization job type enum to rule them all I think, but LGTM!

olacognite (approved):

PR #184, opened 2021-08-18
transformation notifications api client and tests

silvavelosa:
added a new client to create, list and delete transformation notifications.


mathialo (approved):
Nice!

silvavelosa:
Thanks for the review 😄

PR #185, opened 2021-08-18
[MLOPS-739] Modify default values of cpu and memory for Functions

HMEiding:
After we disallowed setting `cpu` and `memory` on Functions in Azure ([here](https://github.com/cognitedata/context-api/pull/2085)), creating a function with the SDK in Azure stopped working. This is because these keyword arguments have default values in the SDK, and so the API returns 400. Also, setting them to `None` does not help, since we don't allow `None` in the API either.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,8 @@ Changes are grouped as follows
 ## [0.59.1]
 
 ### Changed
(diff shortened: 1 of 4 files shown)


alchemistake (approved):

qTipTip (approved):
This looks good to me. I agree on how the discrepancy between API and SDK is maybe not ideal, but I think it should be fine

PR #186, opened 2021-08-26
Add transformation jobs support

silvavelosa:
add support to create jobs (ie. run transformations) and query them

### DIFF
--- a/cognite/experimental/_api/transformations.py
+++ b/cognite/experimental/_api/transformations.py
@@ -1,15 +1,13 @@
-import json as _json
-from typing import List, Optional, Union
+from typing import Awaitable, List, Optional, Union
 
 from cognite.client import utils
 from cognite.client._api_client import APIClient
-from requests import Response
 
 from cognite.experimental._api.transformation_jobs import TransformationJobsAPI
 from cognite.experimental._api.transformation_notifications import TransformationNotificationsAPI
 from cognite.experimental._api.transformation_schedules import TransformationSchedulesAPI
 from cognite.experimental._api.transformation_schema import TransformationSchemaAPI
-from cognite.experimental._constants import HANDLER_FILE_NAME, LIST_LIMIT_CEILING, LIST_LIMIT_DEFAULT, MAX_RETRIES
+from cognite.experimental._constants import LIST_LIMIT_CEILING, LIST_LIMIT_DEFAULT
 from cognite.experimental.data_classes import (
     Transformation,
     TransformationFilter,
@@ -167,28 +165,76 @@ class TransformationsAPI(APIClient):
         """
         return self._update_multiple(items=item)
 
-    def run(self, transformation_id: int) -> TransformationJob:
+    def run(self, transformation_id: int, wait: bool = True) -> TransformationJob:
         """`Run a transformation. <https://docs.cognite.com/api/playground/#operation/runTransformation>`_
 
         Args:
             transformation_id (int): internal Transformation id
+            wait (bool): Wait until the transformation run is finished. Defaults to True.
 
         Returns:
             Created transformation job
 
         Examples:
 
-            Run transformation by id:
+            Run transformation to completion by id:
 
                 >>> from cognite.experimental import CogniteClient
                 >>> c = CogniteClient()
                 >>>
                 >>> res = c.transformations.run(id = 1)
+
+            Start running transformation by id:
+
+                >>> from cognite.experimental import CogniteClient
+                >>> c = CogniteClient()
+                >>>
+                >>> res = c.transformations.run(id = 1, wait = False)
         """
 
-        res = self._post(
+        response = self._post(
             url_path=utils._auxiliary.interpolate_and_url_encode(
                 self._RESOURCE_PATH + "/{}/run", str(transformation_id)
             )
         )
-        return TransformationJob._load(res.json(), cognite_client=self._cognite_client)
+        job = TransformationJob._load(response.json(), cognite_client=self._cognite_client)
+
+        if wait:
+            return job.wait()
+
+        return job
+
+    def run_async(self, transformation_id: int) -> Awaitable[TransformationJob]:
+        """`Run a transformation to completion asynchronously. <https://docs.cognite.com/api/playground/#operation/runTransformation>`_
+
+        Args:
+            transformation_id (int): internal Transformation id
+
+        Returns:
+            Completed transformation job
+
+        Examples:
+
+            Run transformation asyncronously by id:
+
+                >>> import asyncio
+                >>> from cognite.experimental import CogniteClient
+                >>>
+                >>> c = CogniteClient()
+                >>>
+                >>> async def run_transformation():
+                >>>     res = await c.transformations.run_async(id = 1)
+                >>>
+                >>> loop = asyncio.get_event_loop()
+                >>> loop.run_until_complete(run_transformation())
+                >>> loop.close()
+        """
+
+        response = self._post(
+            url_path=utils._auxiliary.interpolate_and_url_encode(
+                self._RESOURCE_PATH + "/{}/run", str(transformation_id)
+            )
+        )
+        job = TransformationJob._load(response.json(), cognite_client=self._cognite_client)
+
+        return job.wait_async()
--- a/cognite/experimental/_api/transformations.py
+++ b/cognite/experimental/_api/transformations.py
@@ -12,6 +12,7 @@ from cognite.experimental._constants import HANDLER_FILE_NAME, LIST_LIMIT_CEILIN
 from cognite.experimental.data_classes import (
     Transformation,
     TransformationFilter,
+    TransformationJob,
     TransformationList,
     TransformationUpdate,
 )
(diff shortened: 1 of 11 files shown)


cognite/experimental/_api/transformations.py
```diff
@@ -164,3 +166,31 @@ def update(
                 >>> res = c.transformations.update(my_update)
         """
         return self._update_multiple(items=item)
+
+    def run(self, transformation_id: int) -> TransformationJob:
```
emelvarol-cognite, line 170: image
Could we have a similar logic here, with `wait` parameter? 🤔

emelvarol-cognite: `async` and `sync` logic is implemented in python sdk for functions as an example so would be very useful to have something like that.

cognite/experimental/data_classes/transformation_jobs.py
```diff
@@ -0,0 +1,125 @@
...
+
+    Args:
+        id (int): A server-generated ID for the object.
+        uuid (UUID): A server-generated UUID for the object.
+        status (TransformationJobStatus): Status of the job.
+        source_project (str): Name of the cdf project where the data will be read from.
+        destination_project (str): Name of the cdf project where the data will be written to.
+        destination_type (str): Target resource type of the transformation.
+        destination_database (str): Target detabase if the destination type is raw.
+        destination_table (str): Target table name if the destination type is raw.
```
hakontro, line 26:
```suggestion
        destination_table (str): Target table name if the destination type is RAW.
```

hakontro (commented):
Awesome :rocket:

cognite/experimental/_api/transformations.py
emelvarol-cognite: Also they cover call by externalId in sdk as it seems, so would be nice to add it until we support it in the backend.

emelvarol-cognite: See the implementation: https://github.com/cognitedata/cognite-sdk-python-experimental/blob/master/cognite/experimental/data_classes/functions.py#L69

emelvarol-cognite: See the implementation in their case:
https://github.com/cognitedata/cognite-sdk-python-experimental/blob/master/cognite/experimental/_api/functions.py#L277

emelvarol-cognite (commented):
Looks great! Please see my comment about `async/sync` support for functions.run  and  run function by `externalId`.

cognite/experimental/data_classes/transformation_jobs.py
```diff
@@ -0,0 +1,125 @@
...
+    def update(self):
+        """`Updates job status. <https://docs.cognite.com/api/playground/#operation/runTransformation>`_"""
+        updated = self.cognite_client.transformations.jobs.retrieve(id=self.id)
+        self.status = updated.status
+        self.error = updated.error
+        self.started_time = updated.started_time
+        self.finished_time = updated.finished_time
+        self.last_seen_time = updated.last_seen_time
+
+    async def completion_coroutine(self, pooling_interval: float = 1):
```
emelvarol-cognite, line 85: I believe this aims to achieve what `wait=True` does. It looks awesome but I think merging this into transformation.run(id, external_id, wait) would be more familiar to the users 🤔 What do you think?

cognite/experimental/_api/transformations.py
silvavelosa: cool! I'm doing something very similar, will add the parameter to run() to keep it consistent.

cognite/experimental/_api/transformations.py
```diff
@@ -164,3 +166,31 @@ def update(
...
+        """
+
+        resource_path = (self._RESOURCE_PATH + f"/{transformation_id}/run",)
+
+        res = self._post(
+            url_path=utils._auxiliary.interpolate_and_url_encode(
+                self._RESOURCE_PATH + "/{}/run", str(transformation_id)
+            )
+        )
+        return TransformationJob._load(res.json(), cognite_client=self._cognite_client)
```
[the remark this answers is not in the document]
silvavelosa: I think the methods with `_` are meant to be internal, not necessarily protected, at least here and in the production SDK (`CogniteResource._load` is used quite frequently from the base `APIClient`). The functions team also has a similar [implementation](https://github.com/cognitedata/cognite-sdk-python-experimental/blob/bb2bdbb860f226807eb2eead7d81cb5e7b760175/cognite/experimental/_api/functions.py#L331).

cognite/experimental/data_classes/transformation_jobs.py
silvavelosa: agree :smiley:

cognite/experimental/_api/transformations.py
silvavelosa: Done :+1:

cognite/experimental/data_classes/transformation_jobs.py
```diff
@@ -0,0 +1,155 @@
...
+                >>> if TransformationJobStatus.FAILED not in [job1.status, job2.status]:
+                >>>     c.transformations.run(id = 3, wait = False)
+        """
+        while self.status not in [TransformationJobStatus.FAILED, TransformationJobStatus.COMPLETED]:
+            time.sleep(1.0)
+            self.update()
+
+        return self
+
+    async def wait_async(self, pooling_interval: float = 1):
```
emelvarol-cognite, line 111: this naming is much better 🚀

cognite/experimental/data_classes/transformation_jobs.py
```diff
@@ -0,0 +1,155 @@
...
+                >>>
+                >>> job1 = c.transformations.run(id = 1, wait = False) 
+                >>> job2 = c.transformations.run(id = 2, wait = False)  
+                >>> job1.wait()
+                >>> job2.wait()
+                >>> if TransformationJobStatus.FAILED not in [job1.status, job2.status]:
+                >>>     c.transformations.run(id = 3, wait = False)
+        """
+        while self.status not in [TransformationJobStatus.FAILED, TransformationJobStatus.COMPLETED]:
+            time.sleep(1.0)
```
emelvarol-cognite, line 106: polling interval?
```suggestion
            time.sleep(polling_interval)
```

cognite/experimental/data_classes/transformation_jobs.py
```diff
@@ -0,0 +1,155 @@
...
+    def update(self):
+        """`Updates job status. <https://docs.cognite.com/api/playground/#operation/runTransformation>`_"""
+        updated = self.cognite_client.transformations.jobs.retrieve(id=self.id)
+        self.status = updated.status
+        self.error = updated.error
+        self.started_time = updated.started_time
+        self.finished_time = updated.finished_time
+        self.last_seen_time = updated.last_seen_time
+
+    def wait(self, pooling_interval: float = 1):
```
emelvarol-cognite, line 86:
```suggestion
    def wait(self, polling_interval: float = 1):
```
ignore  if you meant sth else hehe

cognite/experimental/data_classes/transformation_jobs.py
emelvarol-cognite:
```suggestion
    async def wait_async(self, polling_interval: float = 1):
```

cognite/experimental/data_classes/transformation_jobs.py
silvavelosa: I added the `wait=True` parameter to run. And renamed this to `wait_async` so that users who want to use asyncio can still do it (the implementations are almost the same, the only difference is using the concurrent `asyncio.sleep` instead of the blocking `time.sleep`)

cognite/experimental/_api/transformations.py
```diff
@@ -164,3 +164,77 @@ def update(
                 >>> res = c.transformations.update(my_update)
         """
         return self._update_multiple(items=item)
+
+    def run(self, transformation_id: int, wait: bool = True) -> TransformationJob:
```
emelvarol-cognite, line 168: can you also add `external_id` support as in functions? 🙏

cognite/experimental/data_classes/transformation_jobs.py
```diff
@@ -0,0 +1,155 @@
...
+                >>>
+                >>> async def run_succesive_transformations():
+                >>>     job1 = c.transformations.run(id = 1, wait = False) 
+                >>>     job2 = c.transformations.run(id = 2, wait = False) 
+                >>>     await job1.wait_async()
+                >>>     await job2.wait_async()
+                >>>     if TransformationJobStatus.FAILED not in [job1.status, job2.status]:
+                >>>         c.transformations.run(id = 3, wait = False)
+                >>>
+                >>> ensure_future(run_succesive_transformations())
```
emelvarol-cognite, line 136: awesome docs 🚀

cognite/experimental/data_classes/transformation_jobs.py
silvavelosa: Done, thanks!

cognite/experimental/data_classes/transformation_jobs.py
emelvarol-cognite: great idea 🎸  thanks for providing multiple options 🚀

cognite/experimental/_api/transformations.py
silvavelosa: Done

emelvarol-cognite (approved):
Awesome work!!!!

silvavelosa:
Thanks for the reviews! :rocket:

PR #187, opened 2021-08-27
geospatial v1 experimental support: featuretypes

vincent-cognite:
The business objects defined are the simplest form for now. This is to get started and we can improve when our clients give us feedback.

### DIFF
--- a/cognite/experimental/_api/geospatial.py
+++ b/cognite/experimental/_api/geospatial.py
@@ -1,0 +1,35 @@
+from typing import Any, Dict, List, Union
+
+from cognite.client._api_client import APIClient
+
+from cognite.experimental.data_classes.geospatial import FeatureType
+
+
+class ExperimentalGeospatialAPI(APIClient):
+
+    _RESOURCE_PATH = "/spatial/featuretypes"
+
+    def create_feature_type(
+        self,
+        external_id: str = None,
+        created_time: int = None,
+        last_updated_time: int = None,
+        attributes: Dict[str, Any] = None,
+        cognite_client=None,
+    ) -> FeatureType:
+        """`Creates a feature type`
+
+        Args:
+
+        Returns:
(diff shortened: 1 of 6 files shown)


pawelzwolenskicognite:
what about domain? it looks like SDK works always with default one

cognite/experimental/_api/geospatial.py
```diff
@@ -0,0 +1,103 @@
...
+                ... ]
+                >>> res = c.geospatial.create_feature_types(feature_types)
+
+
+        Args:
+
+        Returns:
+            FeatureType: The created feature types.
+        """
+        return self._create_multiple(items=feature_type)
```
pawelzwolenskicognite, line 45: I'm definitely not a Python guy... where is _create_multiple defined?

pawelzwolenskicognite (approved):
good start!

cognite/experimental/_api/geospatial.py
```diff
@@ -0,0 +1,103 @@
...
+
+        Returns:
+            Union[FeatureType, FeatureTypeList]: Created feature type definition(s)
+
+        Examples:
+
+            Create new type definitions::
+
+                >>> from cognite.experimental import CogniteClient
+                >>> from cognite.experimental.data_classes import Type
```
tuanng-cognite, line 31: import FeatureType, right?

tuanng-cognite (approved):
Very nice!

cognite/experimental/_api/geospatial.py
vincent-cognite: well spotted!

vincent-cognite:
> what about domain? it looks like SDK works always with default one

I was wondering if this should be added now. I was assuming this will be used by APIs sitting on top (and written in Java or Kotlin). Did not want to confuse python user (well, maybe it does anyway)

We can retrofit later easily.

PR #188, opened 2021-08-30
[MLOPS-827] Fail creation of function schedules and calls if retrieving nonce fails

HMEiding:


cognite/experimental/_api/functions.py
```diff
@@ -431,7 +430,7 @@ def _use_token_exchange(cognite_client: CogniteClient):
         nonce = res.json()["items"][0]["nonce"]
         return nonce
     except CogniteAPIError as e:
-        print("Unable to get nonce using token exchange flow. The session API returned with error:", e.message)
+        raise CogniteAPIError("Failed to create session using token exchange flow.") from e
```
alchemistake, line 433: first time I see `raise () from e`

alchemistake (approved):

tests/tests_unit/test_api/test_functions.py
```diff
@@ -625,7 +649,7 @@ def test_list_schedules_with_limit(self, mock_filter_function_schedules_response
     def test_list_schedules_with_function_id_and_function_external_id_raises(self):
         with pytest.raises(AssertionError) as excinfo:
             FUNCTION_SCHEDULES_API.list(function_id=123, function_external_id="my-func")
-            assert "Only function_id or function_external_id allowed when listing schedules." in str(excinfo.value)
+        assert "Only function_id or function_external_id allowed when listing schedules." in str(excinfo.value)
```
qTipTip, line 652: This seems to be indented wrong?

HMEiding: It was indented wrong previously, actually. Having it indented means that it won't be executed. It should be outside the `with` block.

PR #194, opened 2021-09-02
SBS-4034: add support for domains to the geospatial api

vincent-cognite:


cognite/experimental/_api/geospatial.py
```diff
@@ -7,10 +7,20 @@
...
 
+    X_COGNITE_DOMAIN = "x-cognite-domain"
+
     _RESOURCE_PATH = "/spatial/featuretypes"
     _LIST_CLASS = FeatureTypeList
     _ASSERT_CLASSES = False
 
+    _cognite_domain = None
+
+    def set_cognite_domain(self, cognite_domain: str):
```
pawelzwolenskicognite, line 18: how likely is that the domain will be changed? maybe that could be a parameter in constructor?

cognite/experimental/_api/geospatial.py
```diff
@@ -7,10 +7,20 @@
...
+        self._cognite_domain = cognite_domain
+
+    def get_cognite_domain(self):
```
pawelzwolenskicognite, line 21: if we won't go with constructor parameter, maybe it's better to rename that method to "get_current_cognite_domain"?

cognite/experimental/_api/geospatial.py
```diff
@@ -56,7 +66,7 @@ def delete_feature_types(self, external_id: Union[str, List[str]] = None) -> Non
                 >>> c = CogniteClient()
                 >>> c.geospatial.delete_feature_types(external_id=["wells", "pipelines"])
         """
-        self._delete_multiple(external_ids=external_id, wrap_ids=True)
+        return self._with_cognite_domain(lambda: self._delete_multiple(external_ids=external_id, wrap_ids=True))
```
pawelzwolenskicognite, line 69: all that returns look like a good usecase for decorators:
https://realpython.com/primer-on-python-decorators/

cognite/experimental/_api/geospatial.py
vincent-cognite: We don't have direct access to the constructor. It is invoked by CogniteClient and I'm reluctant to put something geospatial specific in the CogniteClient class.

pawelzwolenskicognite: 👍

cognite/experimental/_api/geospatial.py
vincent-cognite: and set_current_cognite_domain()
I was thinking of passing the domain each time in the methods, but thought this could become tedious.

cognite/experimental/_api/geospatial.py
vincent-cognite: So you know a bit of python!

vincent-cognite: The code now has methods whose prototype is (self, *args, **kwargs)
That's real python!

pawelzwolenskicognite (approved):
nice!

PR #199, opened 2021-09-06
rename notification configId to transformationId

silvavelosa:


emelvarol-cognite:
Awesome! We need a version bump hehe :D I will approve after it is done since you cant make changes to approved PR.

emelvarol-cognite (approved):

PR #202, opened 2021-09-10
add transformation_id filter to jobs.list

silvavelosa:


cognite/experimental/_api/transformation_jobs.py
```diff
@@ -27,10 +30,22 @@ def list(self, limit: Optional[int] = LIST_LIMIT_DEFAULT,) -> TransformationJobL
...
+                >>> transformation_jobs_list = c.transformations.jobs.list(transformation_id = 1)
         """
         if limit in [float("inf"), -1, None]:
             limit = LIST_LIMIT_CEILING
 
+        if transformation_id is not None:
+            resource_path = utils._auxiliary.interpolate_and_url_encode(
+                "/transformations/{}/jobs", str(transformation_id)
+            )
+            return self._list(method="GET", limit=limit, resource_path=resource_path)
```
emelvarol-cognite, line 47: could we add external_id support also (as you did in run)🙏

cognite/experimental/_api/transformation_jobs.py
```diff
@@ -10,12 +11,14 @@ class TransformationJobsAPI(APIClient):
...
     _LIST_CLASS = TransformationJobList
 
-    def list(self, limit: Optional[int] = LIST_LIMIT_DEFAULT,) -> TransformationJobList:
+    def list(
+        self, limit: Optional[int] = LIST_LIMIT_DEFAULT, transformation_id: Optional[int] = None
+    ) -> TransformationJobList:
         """`List all running transformation jobs. <https://docs.cognite.com/api/playground/#operation/transformationJobs>`_
 
         Args:
-            cursor (str): Cursor for paging through results.
```
emelvarol-cognite, line 17: why is cursor gone?

emelvarol-cognite: aha makes sense, it is isolated from user

tests/tests_integration/test_api/test_transformation_jobs.py
```diff
@@ -0,0 +1,139 @@
...
+        external_id=f"{prefix}-transformation",
+        destination=TransformationDestination.assets(),
+        query="select id, name from _cdf.assets limit 1000",
+        source_api_key=COGNITE_CLIENT.config.api_key,
+        destination_api_key=COGNITE_CLIENT.config.api_key,
+        ignore_null_fields=True,
+    )
+    ts = COGNITE_CLIENT.transformations.create(transform)
+
+    yield ts
```
emelvarol-cognite, line 26: Do we make a clean-up after? Deleting created transformation,  not sure it is needed though.

cognite/experimental/_api/transformation_jobs.py
silvavelosa: hehe... it was not in the actual method, my mistake was to leave it in documentation.

in the sdk we don't expose the pagination mechanic to the user (there is no cursor returned, nor taken as an input), they only tell us how many items they want max, and we handle the pagination internally

tests/tests_integration/test_api/test_transformation_jobs.py
silvavelosa: nice catch! that actually caused me some trouble testing transformations.list

cognite/experimental/_api/transformation_jobs.py
silvavelosa: this one is tricky... in run, I was able to pull it because run() is in the transformations api class, so I can retrieve a transformation by external_id inside it.

but in this case, list() is in transformation jobs api, class, so retrieving the transformation is not straight forward.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "cognite-sdk-experimental"
-version = "0.60.6"
+version = "0.60.7"
```
emelvarol-cognite, line 3: i'll approve this but the other is merged and you need to bump this version again.

emelvarol-cognite: also i think there are some comments to fix hehe

silvavelosa: thanks 😄 
version bumped and test cleanup fixed... about jobs by transformation external_id, i think it makes sense, but I need to think a little about how to do it, so I'll add it in a next pr

emelvarol-cognite: Awesome!!!! no problem, it was just an idea hehe :D let's merge this one and resolve other stuff instead.

emelvarol-cognite: I think a test needs to be fixed before i approve.

emelvarol-cognite (approved):

silvavelosa:
Thanks for the review! 😃

PR #209, opened 2021-09-22
Update functions/call docs + minor cleanup

haakonvt:
Tiny stuff that doesnt warrant a new release imo. Let me know if you think otherwise


pavelzubarev (approved):
Nice ones!

PR #211, opened 2021-09-23
SBS-4070: add support for the attribute selection

vincent-cognite:


cognite/experimental/_api/geospatial.py
```diff
@@ -266,11 +269,18 @@ def search_features(self, feature_type: FeatureType, filter: Dict[str, Any], lim
                 >>> res = c.geospatial.search_features(my_feature_type, filter={"range": {"attribute": "temperature", "gt": 12.0}})
                 >>> for f in res:
                 ...     # do something with the features
+
+            Search for features and select output attributes:
+
+                >>> res = c.geospatial.search_features(my_feature_type, filter={}, attributes={"temperature": {}, "pressure": {}})
```
tuanng-cognite, line 275: this is a little bit awkward to use this output definition: What can be inside "temperature": {}. A clearer example is desired.

vincent-cognite: :) I agree, but the issue is that I cannot give a better example. The {} is for the future. Geometry format selection is coming soon, otherwise, potential unit conversions.
I will give a better example when geometry format is implemented.

PR #212, opened 2021-09-24
Remove unneeded raw_type parameter

silvavelosa:
`TransformationDestination.raw()` now uses `type = 'raw'` istead of `'raw_table'` and doesn't use deprecated `raw_type` parameter.


emelvarol-cognite (approved):
A W E S O M E

PR #217, opened 2021-09-27
Add optional timeout to run, wait and async

silvavelosa:


cognite/experimental/data_classes/transformation_jobs.py
```diff
@@ -137,21 +141,40 @@ def wait(self, polling_interval: float = 1):
...
                 >>>     c.transformations.run(id = 3, wait = False)
+
+            wait transformation for 5 minutes and do something if still running:
+
+                >>> from cognite.experimental import CogniteClient
+                >>> c = CogniteClient()
+                >>>
+                >>> job = c.transformations.run(id = 1, wait = False)  
+                >>> if job.wait(timeout = 5.0*60).status not in [TransformationJobStatus.FAILED, TransformationJobStatus.COMPLETED]:
+                >>>     # send notification of long running transformation
```
emelvarol-cognite, line 152:
```suggestion
                >>>     # send notification of long running transformation
```
is this leftover from some other comment?

silvavelosa: It was intended to mean like "here goes the code of whatever you want to do if the transformation is still running, for example, sending a notification".... but didn't find a clear way to put it, other than a comment.

silvavelosa: changed it a bit, maybe it's more instructive now

emelvarol-cognite: thanks! it makes sense, sorry for missing the point :D

CHANGELOG.md
```diff
@@ -12,6 +12,11 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [0.60.15]
```
emelvarol-cognite, line 15: I am thinking if we are increasing versions unnecessarily, it doesn't publish the new stuff when we don't bump it, right? that was my impression but i am not sure as i see some people do not bump it from time to time.

emelvarol-cognite (approved):
superr

CHANGELOG.md
silvavelosa: when the version isn't pushed, the release to PyPI step in the release action [throws an error](https://github.com/cognitedata/cognite-sdk-python-experimental/runs/3719961347?check_suite_focus=true#step:10:35).

I think that response means it doesn't actually get published

PR #218, opened 2021-10-01
fix streaming

pawelzwolenskicognite:
the bug introduced by moving streaming format to output spec
Isn't that tested that we didn't spot it before?


pawelzwolenskicognite:
@vincent-cognite can I ask you to fix the format of the files? I still have troubles with setting up python locally and I doubt I will be able to do that before my vacations

vincent-cognite:
> sorry, my mistake

no worries!

PR #220, opened 2021-10-01
update github workflow to take OIDC config

chingyulin:
Skipping some transformations and fuctions' tests as communicated.

### DIFF
--- a/tests/tests_integration/test_api/test_transformation_jobs.py
+++ b/tests/tests_integration/test_api/test_transformation_jobs.py
@@ -63,9 +63,9 @@ async def other_running_transformation(other_transformation):
         yield transform
 
 
+@pytest.mark.skip(reason="Not compatible with tokens, data integration team will follow up.")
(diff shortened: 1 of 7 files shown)


tests/tests_integration/test_api/test_transformation_jobs.py
```diff
@@ -65,6 +65,7 @@ async def other_running_transformation(other_transformation):
 
 class TestTransformationJobsAPI:
     @pytest.mark.asyncio
+    @pytest.mark.skip(reason="Not compatible with tokens, data integration team will follow up.")
```
erlendvollset, line 68: You can place skip on the test class rather than each individual test

chingyulin: place the skip on the transformation test classes; keep skip in functions on the method level as only a subset is failing

olacognite (approved):
Looks amazingly super greatest to me

PR #221, opened 2021-10-04
Add query

silvavelosa:
Add `transformationsAPI.preview()` method to run a query and get a preview of the results


emelvarol-cognite (approved):
A W E S O M E! version bump pls, and also can you add a new DataSource type for DataSets? type="data_sets"

PR #224, opened 2021-10-06
Cxt 292

chingyulin:


tests/tests_integration/test_api/test_geospatial.py
```diff
@@ -211,6 +211,7 @@ def test_retrieve_multiple_feature_types_by_external_id(
         )
         assert COGNITE_CLIENT.geospatial.get_current_cognite_domain() == cognite_domain
 
+    @pytest.mark.skip(reason="This test fails several times.")
```
chingyulin, line 214: @vincent-cognite FYI

cyriaquebrousse (approved):
Nice!
What is the plan for the integration tests?

chingyulin:
> Nice! What is the plan for the integration tests?

Created a ticket for it. https://cognitedata.atlassian.net/browse/CXT-375
I didn't implement it here because we don't have a base template to test against in the test project. It should've been introduced when the schema completion endpoint was added.

PR #225, opened 2021-10-06
enable request gzip again now that coregis server supports it 

vincent-cognite:


tuanng-cognite:
Can we refer to what change in coregis server to make it support gzip in request body? I'm not aware of.

vincent-cognite:
@tuanng-cognite 
it was https://github.com/cognitedata/coregis/pull/399

PR #235, opened 2021-11-01
fix data classes to string circular reference

silvavelosa:
many transformation data classes defined `cls.cognite_client` instead of `cls._cognite_client`. This caused a circular reference exception when dumping them to json on `str(cls)`


emelvarol-cognite (approved):
good catch!!!! update the version plsss

PR #242, opened 2021-11-22
point transformations to v1 api

silvavelosa:
transformations api is now avalable in v1, so playground users will use that version now

### DIFF
--- a/cognite/experimental/_client.py
+++ b/cognite/experimental/_client.py
@@ -112,7 +112,7 @@ class CogniteClient(Client):
             self._config, api_version="playground", cognite_client=self
         )
 
-        self.transformations = TransformationsAPI(self._config, api_version="playground", cognite_client=self)
+        self.transformations = TransformationsAPI(self._config, api_version="v1", cognite_client=self)
 
         self.diagrams = DiagramsAPI(self._config, api_version=self._API_VERSION, cognite_client=self)
         # template completion only


emelvarol-cognite:
I thought `cognite-sdk-python` repo would use v1 and this would keep going with `playground`? 🤔

emelvarol-cognite:
Let's see if tests will pass on v1 first :D

silvavelosa:
> 

🤔 I was following what I see Geospatial is doing.... but you're probably right

emelvarol-cognite:
> > 
> 
> 🤔 I was following what I see Geospatial is doing.... but you're probably right

That also makes sense, maybe we want to kill the playground? @hakontro  what do you think?

hakontro:
Not sure what Geospatial is doing, but I agree with Emel: We want this on the main SDK, not the experimental one. Eventually we will delete the code here from the experimental SDK, and shut down playground, but there's no immediate rush.

hakontro:
I would however guess that v1 endpoints shouldn't be in this experimental SDK at all, are there others on v1 here @silvavelosa ?

silvavelosa:
> I would however guess that v1 endpoints shouldn't be in this experimental SDK at all, are there others on v1 here @silvavelosa ?

yup, Geospatials API is pointing to v1

emelvarol-cognite:
> > I would however guess that v1 endpoints shouldn't be in this experimental SDK at all, are there others on v1 here @silvavelosa ?
> 
> yup, Geospatials API is pointing to v1

Maybe they wanted to deprecate playground API without deprecating experimental SDK, could be a specific case. 🤔

hakontro:
Not doing this one! :)

PR #243, opened 2021-11-22
support orderBy in geospatial sdk

pawelzwolenskicognite:


cognite/experimental/_api/geospatial.py
```diff
@@ -306,12 +312,17 @@ def search_features(
...
+                >>> res = c.geospatial.search_features(my_feature_type, filter={}, orderBy={"temperature": "ASC", "pressure": "DESC"})
+
         """
         resource_path = self._feature_resource_path(feature_type) + "/search"
         cls = FeatureList
         resource_path = resource_path
         res = self._post(
-            url_path=resource_path, json={"filter": filter, "limit": limit, "output": {"attributes": attributes}}
+            url_path=resource_path,
+            json={"filter": filter, "limit": limit, "output": {"attributes": attributes}, "orderBy": orderBy},
```
vincent-cognite, line 325: since "attributes" and "orderBy" are fixed names, we should probably introduce a class for it

pawelzwolenskicognite: where do we keep that classes? as inner classes in `ExperimentalGeospatialAPI` ?

vincent-cognite: that would be in cognite/experimental/data_classes/geospatial.py

vincent-cognite (approved):

PR #245, opened 2021-11-30
[CDF-13694]Python updates according to changes in API. 

AlexThunder:
Add notification_config field to extPipe entity.
Add filtering possibilities to ep.list().


haakonvt:
Closed due to repo being archived.

PR #248, opened 2021-12-01
add geospatial aggregate

pawelzwolenskicognite:


cognite/experimental/data_classes/geospatial.py
```diff
@@ -6,8 +6,7 @@
 
 
 class FeatureType(CogniteResource):
-    """A representation of a feature type in the geospatial api.
-    """
+    """A representation of a feature type in the geospatial api."""
```
pawelzwolenskicognite, line 9: reformatting is due to running `black` command locally

tests/tests_integration/test_api/test_geospatial.py
```diff
@@ -421,3 +421,27 @@ def test_from_geopandas(self, test_feature_type, two_test_features):
...
+            aggregates=["count"],
+        )
+        assert res[0].count == 1
+
+        res = COGNITE_CLIENT.geospatial.aggregate_features(
+            feature_type=test_feature_type,
+            filter={},
+            attribute="temperature",
+            aggregates=["min", "max"],
+            group_by=["externalId"],
```
vincent-cognite, line 445: dubious grouping 🤔

vincent-cognite: maybe we need a few more features to test the grouping with a "groupable" attribute?

vincent-cognite (approved):
good enough, maybe improve the test a bit?

tests/tests_integration/test_api/test_geospatial.py
pawelzwolenskicognite: I disagree on adding more tests or changing them. This is not a Python SDK responsibility to test API. The only thing that has to be tested here is if SDK uses given parameters in http call.

Generally if that was me, I would never write such tests. They call real API which is problematic (look that some of the builds above failed... because of some other API being not responsive). I would rather go with contract testing (https://docs.pact.io/) and generated service stubs but since this is a shared repository I won't push for that.

vincent-cognite: but you are not really testing that the group_by has been applied here, since you only check the number of responses

pawelzwolenskicognite: I called method without groupBy and I got 1 items, then I called it with groupBy and I got 2 items. I think it is clear that groupBy is applied

PR #250, opened 2021-12-06
allow list jobs by external id

silvavelosa:
move jobs.list filter to query params

### DIFF
--- a/cognite/experimental/_api/transformation_jobs.py
+++ b/cognite/experimental/_api/transformation_jobs.py
@@ -4,7 +4,12 @@ from cognite.client import utils
 from cognite.client._api_client import APIClient
 
 from cognite.experimental._constants import LIST_LIMIT_CEILING, LIST_LIMIT_DEFAULT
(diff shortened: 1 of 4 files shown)


cognite/experimental/_api/transformation_jobs.py
```diff
@@ -40,13 +49,9 @@ def list(
         if limit in [float("inf"), -1, None]:
             limit = LIST_LIMIT_CEILING
 
-        if transformation_id is not None:
-            resource_path = utils._auxiliary.interpolate_and_url_encode(
-                "/transformations/{}/jobs", str(transformation_id)
-            )
-            return self._list(method="GET", limit=limit, resource_path=resource_path)
+        filter = TransformationJobsFilter(transformation_id=transformation_id).dump(camel_case=True)
```
emelvarol-cognite, line 52: here, we also need to include transformation_external_id 👯

silvavelosa: Nice catch 👍

emelvarol-cognite (approved):

PR #259, opened 2022-01-05
[MLOPS-951] Increase limits in doc strings for envvars and secrets

qTipTip:
https://github.com/cognitedata/context-api/pull/2473
- Update docstring to reflect increased limits on secrets and env-vars.
- Bump version and add changelog.


HMEiding (changes requested):
I believe the docs are always updated no matter if there is a version bump or not. And since we're not doing any code changes here, maybe we should skip the bump?

HMEiding (approved):
LGTM

PR #260, opened 2022-01-05
[MLOPS-264] Functions: Overwrite file in filesAPI

HMEiding:
When creating a function with `function_handle` or `folder` and with `external_id` specified, the zip file uploaded to the Files API is given an external-id and is overwritten if it already exists.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [0.60.37]
+
+### Changed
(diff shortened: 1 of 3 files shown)


qTipTip:
Is there any occasion where we *want* to deploy a function with `external_id`, but without overwriting?

qTipTip (approved):
Looks good to me! :)

HMEiding:
> Is there any occasion where we _want_ to deploy a function with `external_id`, but without overwriting?

In the case where the user actually wants to keep the file in the files-API, I believe the user would use the `file_id` argument instead of `folder` or `function_handle`.

PR #262, opened 2022-01-12
[mlops-974] Use ast instead of a module import to check handler

anothercoginthewheel:
Use ast to check for the presence of handle function and validate if the arguments accepted by this function.


haakonvt:
Closed due to repo being archived.

PR #263, opened 2022-01-12
Remove transformations from experimental

silvavelosa:
since transformations is now available in v1 and supported on the official sdk, the code is no longer needed in the experimentals sdk


emelvarol-cognite (approved):

emelvarol-cognite:
version?

PR #264, opened 2022-01-14
Cxt 460 : Support for AnnotationsV2

peetcremer1:
Support for AnnotationsV2 in the experimental python SDK

For now, the API Client for them lives aside the one for AnnotationsV1
The legacy annotations are accessed with `client.annotations`
AnnotationsV2 are accessed with `client.annotations_v2`


cognite/experimental/_api/annotations_v2.py
```diff
@@ -0,0 +1,105 @@
...
+            filter["annotatedResourceIds"] = [
+                {to_camel_case(k): v for k, v in f.items()} for f in filter["annotatedResourceIds"]
+            ]
+
+        return self._list(method="POST", limit=limit, filter=filter)
+
+    @staticmethod
+    def _convert_resource_to_patch_object(resource: CogniteResource, update_attributes: List[str]):
+        if not isinstance(resource, AnnotationV2):
+            return APIClient._convert_resource_to_patch_object(resource, update_attributes)
```
chingyulin, line 58: why is this necessary?

cognite/experimental/data_classes/annotations_v2.py
```diff
@@ -0,0 +1,229 @@
...
+        self.creating_app = creating_app
+        self.creating_app_version = creating_app_version
+        self.creating_user = creating_user
+        self.annotated_resource_type = annotated_resource_type
+        self.annotated_resource_id = annotated_resource_id
+        self.annotated_resource_external_id = annotated_resource_external_id
+        self.linked_resource_id = linked_resource_id
+        self.linked_resource_external_id = linked_resource_external_id
+        self.linked_resource_type = linked_resource_type
+        self.id = None  # Read only
```
chingyulin, line 69: What do you mean by Read only? Is it guaranteed somehow?

cognite/experimental/data_classes/annotations_v2.py
```diff
@@ -0,0 +1,229 @@
...
+        created_time (int, optional): Time since this annotation was created in CDF. The time is measured in milliseconds since 00:00:00 Thursday, 1 January 1970, Coordinated Universal Time (UTC), minus leap seconds. Read only.
+        last_updated_time (int, optional): Time since this annotation was last updated in CDF. The time is measured in milliseconds since 00:00:00 Thursday, 1 January 1970, Coordinated Universal Time (UTC), minus leap seconds. Read only.
+        
+        cognite_client (CogniteClient, optional): The client to associate with this object. Read only.
+    """
+
+    def __init__(
+        self,
+        annotation_type: str,
+        data: dict,
```
cyriaquebrousse, line 45: Here we could use the `annotation-types` directly, since they are also in Python?

peetcremer1: Yes we could. However I opted against integrating with the annotation-types repo for now and only use python buildin types to keep the implementation simple and the dependencies minimal. In a future version, when the annotation-types package is published publicly, the constructor should take the enum defined in there

cognite/experimental/data_classes/annotations_v2.py
peetcremer1: With "Read only" I mean that this field is not supposed to be filled by the user, only by the API (which calls the _load _load function when creating the annotation on the API, for example). For the user, those "Read only" fields are simply not offered it as an argument to the constructor. This is no strong guarantee of course, because the user could simply change the `.id` property outside of the constructor if they really want to.

One could make it harder by hiding `.id` behind an `@property` getter without a setter and use an internal property `._id`. Still, there would be nothing stopping the user from manually setting the internal `._id` property

cognite/experimental/_api/annotations_v2.py
peetcremer1: There is a `_convert_resource_to_patch_object` function in the base class, the purpose of which is to extract the updatable fields from a dataclass instance (e.g. `AnnotationV2`) and create the update dictionary from it before the `update` call to the API.

The `creatingUser` field needs some custom handling because of its special `None` value. By default, all `None` values are filtered out of the update dictionary. 
The `AnnotationV2Update` dataclass already implements custom handling of `creatingUser`. So the easiest way to get the update dictionary is to create a `AnnotationV2Update` and dump it.

peetcremer1:
I just refactored the integration tests quite a bit to make the cleanup more robust.

cognite/experimental/_api/annotations_v2.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def delete(self, id: Union[int, List[int]]) -> None:
+        """Delete annotations
+
+        Args:
+            id (Union[int, List[int]]): ID or list of IDs to be deleted
+        """
+        self._delete_multiple(ids=id, wrap_ids=True)
+
+    def retrieve_multiple(self, ids: List[int]) -> AnnotationV2List:
```
cyriaquebrousse, line 88: Could we merge this with `retrieve`, similarly to what is done for `delete`?

cognite/experimental/data_classes/annotations_v2.py
```diff
@@ -0,0 +1,229 @@
...
+
+        linked_resource_type (str, optional): The CDF resource type of an optional linked CDF resource. Defaults to None.
+        linked_resource_id (int, optional): The server-generated id of an optional linked CDF resource. Defaults to None.
+        linked_resource_external_id (str, optional): The user-defined id of an optional linked CDF resource. Defaults to None.
+
+        id (int, optional): A server-generated id for the object. Read only.
+        created_time (int, optional): Time since this annotation was created in CDF. The time is measured in milliseconds since 00:00:00 Thursday, 1 January 1970, Coordinated Universal Time (UTC), minus leap seconds. Read only.
+        last_updated_time (int, optional): Time since this annotation was last updated in CDF. The time is measured in milliseconds since 00:00:00 Thursday, 1 January 1970, Coordinated Universal Time (UTC), minus leap seconds. Read only.
+        
+        cognite_client (CogniteClient, optional): The client to associate with this object. Read only.
```
cyriaquebrousse, line 39:
```suggestion
        annotation_type (str): The type of the annotation. This uniquely decides what the structure of the 'data' block will be.
        data (dict): The annotation information. The format of this object is decided by and validated against the 'annotation_type' attribute.
        status (str): The status of the annotation, e.g. "suggested", "approved", "rejected".
        
        annotated_resource_type (str): Type name of the CDF resource that is annotated, e.g. "file".
        annotated_resource_id (int, optional): The internal ID of the annotated resource. Either this field or the external ID of the linked resource is required (not both).
        annotated_resource_external_id (str, optional): The external ID of the annotated resource. Either this field or the internal ID of the linked resource is required (not both).

        creating_app (str): The name of the app from which this annotation was created.
        creating_app_version (str): The version of the app that created this annotation. Must be a valid semantic versioning (SemVer) string.
        creating_user: (str, optional): A username, or email, or name. This is not checked nor enforced. If the value is None, it means the annotation was created by a service.

        linked_resource_type (str, optional): The CDF resource type of an optional linked CDF resource. Defaults to None.
        linked_resource_id (int, optional): The internal ID of the linked resource. Either this field or the external ID of the linked resource can be provided (not both).
        linked_resource_external_id (str, optional): The external ID of the linked resource. Either this field or the internal ID of the linked resource can be provided (not both).

        id (int, optional): A server-generated id for the object. Read-only.
        created_time (int, optional): Time when this annotation was created in CDF. The time is measured in milliseconds since 00:00:00 Thursday, 1 January 1970, Coordinated Universal Time (UTC), minus leap seconds. Read-only.
        last_updated_time (int, optional): Time when this annotation was last updated in CDF. The time is measured in milliseconds since 00:00:00 Thursday, 1 January 1970, Coordinated Universal Time (UTC), minus leap seconds. Read-only.
        
        cognite_client (CogniteClient, optional): The client to associate with this object. Read-only.
```
Can we stay close to the documentation as provided in the service contracts? I think if we want to change something it should be done there. Someday™ we will generate the SDK docs automatically from the service contracts.

cognite/experimental/data_classes/annotations_v2.py
```diff
@@ -0,0 +1,229 @@
...
+        self.annotated_resource_type = annotated_resource_type
+        self.annotated_resource_id = annotated_resource_id
+        self.annotated_resource_external_id = annotated_resource_external_id
+        self.linked_resource_id = linked_resource_id
+        self.linked_resource_external_id = linked_resource_external_id
+        self.linked_resource_type = linked_resource_type
+        self.id = None  # Read only
+        self.created_time = None  # Read only
+        self.last_updated_time = None  # Read only
+        self._cognite_client = None  # Read only
```
cyriaquebrousse, line 72: Curious to when this is used? Is it for the user's convenience?

cognite/experimental/data_classes/annotations_v2.py
peetcremer1: Yes, I like it! I tried to stay close to the docstrings of the core CDF resources like Events but I like staying close to the contracts more. As you say, the docs will then not change too drastically when someday™ we can autogenerated them.

cognite/experimental/data_classes/annotations_v2.py
peetcremer1: Currently it is not used at all. All other resources have it though and use it to implement convenience functions on the dataclass itself. In this way, users could, e.g., fetch an annotation, make some changes to fields and just call annotation.update(). Nothing of that is implemented yet, but I could see it become useful down the line.

Also the `_load` function implicitly expects there to be a `self._cognite_client` variable in place, so not having it would actually mean more work.

cognite/experimental/_api/annotations_v2.py
peetcremer1: We could, but it would not be in the spirit of the existing APIs.
I don't know the reason for it, but the APIs for assets, events, ... all have seperate `retrieve` and `retrieve_multiple` methods, but merged `delete`, `create` and `update` methods.

cyriaquebrousse: Okay, I checked the Assets API and they do the same there, for some obscure reason.

cognite/experimental/data_classes/annotations_v2.py
```diff
@@ -0,0 +1,229 @@
...
+        self.creating_app = creating_app
+        self.creating_app_version = creating_app_version
+        self.linked_resource_type = linked_resource_type
+        self.linked_resource_ids = linked_resource_ids
+        self.annotation_type = annotation_type
+        self._cognite_client = (
+            None  # Read only. Will be filled by superclass load. # TODO is this true? Will this ever be used?
+        )
+
+    def dump(self, camel_case: bool = False):
```
cyriaquebrousse, line 155: Do we need to override load() for handling creating_user properly?

peetcremer1: The base class `CogniteFilter` does not actually have a `_load()` method, so there is nothing to override. For some reason, the `AssetFilter` which is derived from `CogniteFilter` does override `_load()` but I think it is never called and I can't see how it would work.

tests/tests_integration/test_contextualization/test_annotations_v2.py
```diff
@@ -0,0 +1,221 @@
...
+
+import pytest
+from cognite.client.exceptions import CogniteAPIError
+
+from cognite.experimental import CogniteClient
+from cognite.experimental.data_classes import AnnotationV2, AnnotationV2Filter, AnnotationV2List, AnnotationV2Update
+
+COGNITE_CLIENT = CogniteClient()
+ANNOTATIONSAPI = COGNITE_CLIENT.annotations_v2
+
```
cyriaquebrousse, line 12: Those can be fixtures

tests/tests_integration/test_contextualization/test_annotations_v2.py
```diff
@@ -0,0 +1,221 @@
...
+def delete_with_check(delete_ids: List[int], check_ids: Optional[List[int]] = None) -> None:
+    if check_ids is None:
+        check_ids = delete_ids
+    ANNOTATIONSAPI.delete(id=delete_ids)
+    try:
+        ANNOTATIONSAPI.retrieve_multiple(check_ids)
+        raise ValueError(f"retrieve_multiple after delete successful for ids {check_ids}")
+    except CogniteAPIError as e:
+        assert "Some annotation(s) not found" in str(e)
```
cyriaquebrousse, line 22: This is a bit too detailed. We should be able to change the return string in the API without it being a breaking change for the SDK CI/CD pipeline. The following assertions are good though.

tests/tests_integration/test_contextualization/test_annotations_v2.py
```diff
@@ -0,0 +1,221 @@
...
+        annotated_resource_ids=[{"id": fid} for fid in file_ids],
+        creating_app="UnitTest",
+    )
+    annotation_ids = [a.id for a in ANNOTATIONSAPI.list(filter=filter)]
+    if annotation_ids:
+        delete_with_check(annotation_ids, check_ids)
+
+
+@pytest.fixture
+def file_id() -> int:
```
cyriaquebrousse, line 42: How about creating a file upon startup of your suite, running your tests (creating annotations), and then on teardown, delete all created annotations and the file itself?
This would be more robust and independent from what's actually in the test project.

peetcremer1: Yes, this is probably better. The above is the lazy and less robust way which I just took over from the old tests.

tests/tests_integration/test_contextualization/test_annotations_v2.py
```diff
@@ -0,0 +1,232 @@
...
+        creating_user=annotation.creating_user,
+    )
+    annotations_list = ANNOTATIONSAPI.list(filter=filter, limit=limit)
+    assert isinstance(annotations_list, AnnotationV2List)
+    if limit == -1 or limit > len(annotations):
+        assert len(annotations_list) == len(annotations)
+    else:
+        assert len(annotations_list) == limit
+
+    # TODO can use check_created_vs_base here?
```
cyriaquebrousse, line 95: Leftover? Or do you want to do this in this PR?

peetcremer1:
Further simplified the tests with fixtures, removed the try / finally guards everywhere, and added the entry to the CHANGELOG.md. Anything more to do?

cyriaquebrousse (approved):
🚢

PR #270, opened 2022-02-03
vision: add endpoints for asset detection

andersfylling:


andersfylling:
I don't think that `items` in the response should be required: https://docs.cognite.com/api/playground/#operation/visionTagDetectionRetrieve

cognite/experimental/_api/vision.py
```diff
@@ -0,0 +1,58 @@
...
+    DetectAssetsInFilesJob,
+    EitherFileId,
+    InternalId,
+)
+from cognite.experimental.utils import resource_to_camel_case
+
+
+class VisionAPI(APIClient):
+    _RESOURCE_PATH = "/context/vision"
+    _TAG_DETECTION_PATH = _RESOURCE_PATH + "/tagdetection"
```
marvrez, line 16: Nit pick: maybe use string-literals here instead? E.g. `f"{_RESOURCE_PATH}/tagdetection"`

cognite/experimental/data_classes/vision.py
```diff
@@ -0,0 +1,237 @@
...
+@dataclass
+class AllOfFileId(InternalFileId):
+    file_external_id: Optional[ExternalId] = None
+
+
+class JobStatus(str, Enum):
+    QUEUED = "Queued"
+    RUNNING = "Running"
+    COMPLETED = "Completed"
+    FAILED = "Failed"
```
marvrez, line 36: `JobStatus` already exists [here](https://github.com/cognitedata/cognite-sdk-python/blob/45223379f5c78bd2c2dc78224cb5aba132149a27/cognite/client/data_classes/contextualization.py#L15), and has the same usage, so I guess we can import it from contextualization.

cognite/experimental/data_classes/vision.py
```diff
@@ -0,0 +1,237 @@
...
+            )
+            items = k.get("items")
+            if items is not None:
+                instance.items = [
+                    AllOfFileId(file_id=item["file_id"], file_external_id=item.get("file_external_id"))
+                    for item in k["items"]
+                ]
+            return instance
+
+        raise TypeError("Resource must be json str or Dict, not {}".format(type(resource)))
```
marvrez, line 89: Nit pick: string literals would make the code here more explicit and compact. E.g., `f"Resource must be json str or Dict, not {type(resource)}"`

cognite/experimental/data_classes/vision.py
```diff
@@ -0,0 +1,237 @@
...
+            return cls._load(json.loads(resource))
+        elif isinstance(resource, Dict):
+            k = resource_to_snake_case(resource)
+            return cls(
+                error_message=k["error_message"],
+                items=[
+                    AllOfFileId(file_id=v["file_id"], file_external_id=v.get("file_external_id")) for v in k["items"]
+                ],
+            )
+        raise TypeError("Resource must be json str or Dict, not {}".format(type(resource)))
```
marvrez, line 111: Same here and for the rest of the data classes.

tests/tests_integration/test_vision/test_asset_detection.py
```diff
@@ -0,0 +1,49 @@
...
+from cognite.client.data_classes import FileMetadata
+from cognite.client.exceptions import CogniteAPIError
+
+from cognite.experimental import CogniteClient
+from cognite.experimental.data_classes import (
+    EntityMatchingMatchList,
+    EntityMatchingPipeline,
+    EntityMatchingPipelineRunList,
+    EntityMatchingPipelineUpdate,
+)
```
marvrez, line 11: These ones as well

tests/tests_integration/test_vision/test_asset_detection.py
```diff
@@ -0,0 +1,49 @@
+import pytest
+from cognite.client.data_classes import FileMetadata
+from cognite.client.exceptions import CogniteAPIError
```
marvrez, line 3: This import seems to be unused

tests/tests_unit/test_vision/test_asset_detection.py
```diff
@@ -0,0 +1,92 @@
...
+        "createdTime": 934875934785,
+        "startTime": 934875934785,
+        "statusTime": 934875934785,
+        "jobId": 1,
+        "items": [
+            {"fileId": 1},
+            {"fileId": 2, "fileExternalId": "some_external_id"},
+            {"fileId": 3},
+            {"fileId": 4},
+            {"fileId": 5, "fileExternalId": "another"},
```
marvrez, line 25: For completeness, the parameters to the tag detection API should also be in the job response body. Namely, `partialMatch` and `useCache` and `assetSubtreeIds` should also be in this dict.

tests/tests_unit/test_vision/test_asset_detection.py
```diff
@@ -0,0 +1,92 @@
...
+
+@pytest.fixture
+def mock_fetch_job_ok(rsps):
+    response_body = {
+        "status": "Completed",
+        "createdTime": 934875934785,
+        "startTime": 934875934785,
+        "statusTime": 934875934785,
+        "jobId": 1,
+        "items": [{"fileId": 1, "annotations": [{"text": "testing", "assetIds": [1, 2, 3], "confidence": 0.9}],},],
```
marvrez, line 42: Same here.

tests/tests_unit/test_vision/test_asset_detection.py
```diff
@@ -0,0 +1,92 @@
...
+        assert contains_file_id(job.items, id=2), job
+        assert contains_file_id(job.items, id=3), job
+        assert contains_file_id(job.items, external_id="another"), job
+
+    def test_retrieve_job(self, mock_fetch_job_ok):
+        job = VAPI.retrieve_detected_assets_in_files_job(job_id=1)
+        assert isinstance(job, DetectAssetsInFilesJob), "wrong instance returned"
+        assert job.status == "Completed", job
+        assert job.job_id == 1, job
+        assert contains_file_id(job.items, id=1), job
```
marvrez, line 92: Perhaps we can also add a small test for checking that the `items` entry exists for this mock job? `items` should always exist in the returned json when the status for a job is `"Completed"`.

marvrez (commented):
Really great work! I added a few comments with some rather minor suggestions.

cognite/experimental/data_classes/vision.py
marvrez: Or, we could at least sub-class from their enum class. Since if they add/remove a job status, we also have to make a corresponding change.

marvrez:
> I don't think that items in the response should be required: https://docs.cognite.com/api/playground/#operation/visionTagDetectionRetrieve

`items` is always included in the response (even if no tags were detected in the files) when the status is `"Completed"`

andersfylling:
> > I don't think that items in the response should be required: [docs.cognite.com/api/playground/#operation/visionTagDetectionRetrieve](https://docs.cognite.com/api/playground/#operation/visionTagDetectionRetrieve)
> 
> `items` is always included in the response (even if no tags were detected in the files) when the status is `"Completed"`

interesting. I seem to get an error on my integration test, hence why I made it Optional.
image

andersfylling:
I chose to not inherit from the ContextAPI as we have different requirements for json transformation. I see that p&id detect method allows a dictionary of custom json keys. Using the same transformation as vision requires will potentially break their internal behavior.

marvrez (approved):
Great job! LGTM

PR #272, opened 2022-02-08
[MLOPS-996] Functions added Extra Index URLs

alchemistake:


CHANGELOG.md
```diff
@@ -12,6 +12,12 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [0.66]
+
+### Added
+
+- Optional argument `extraIndexUrls` to `FunctionsAPI.create` which adds extra index URLs to python package manager.
```
HMEiding, line 19:
```suggestion
- Optional argument `extraIndexUrls` to `FunctionsAPI.create` which adds extra index URLs to python package manager. This enables the user to install pip packages from private repositories.
```

HMEiding (approved):
LGTM 🚀 
Just one small suggestion.

CHANGELOG.md
alchemistake: I think is just repeating the same point, we might merge the sentences directly.

cognite/experimental/_api/functions.py
```diff
@@ -84,7 +86,8 @@ def create(
             memory (Number, optional):              Memory per function measured in GB. Allowed values are in the range [0.1, 2.5], and None translates to the API default which is 1 GB in GCP. The argument is unavailable in Azure.
             runtime (str, optional):                The function runtime. Valid values are ["py37", "py38", "py39", `None`], and `None` translates to the API default which currently is "py38". The runtime "py3x" resolves to the latest version of the Python 3.x.y series.
             metadata (Dict[str, str], optional):    Metadata for the function as key/value pairs. Key & values can be at most 32, 512 characters long respectively. You can have at the most 16 key-value pairs, with a maximum size of 512 bytes.
-
+            extra_index_urls (List[str], optional):   Extra Index URLs for Python Package Manager to use. Please do not use any extra indices you don't trust. The authentication provided in the URL should not have right access. If you are not careful, these can cause [supply chain attack](https://en.wikipedia.org/wiki/Supply_chain_attack)
```
qTipTip, line 89:
```suggestion
            extra_index_urls (List[str], optional):   Extra Index URLs for Python Package Manager to use. Please do not use any extra indices you don't trust. The authentication provided in the URL should not have write-access. If you are not careful, these can cause [supply chain attack](https://en.wikipedia.org/wiki/Supply_chain_attack)
```

cognite/experimental/_api/functions.py
```diff
@@ -84,7 +86,8 @@ def create(
...
+            index_url (str, optional):               Index URL for Python Package Manager to use. Please do not use any indices you don't trust. The authentication provided in the URL should not have right access. If you are not careful, these can cause [supply chain attack](https://en.wikipedia.org/wiki/Supply_chain_attack)
```
qTipTip, line 90:
```suggestion
            index_url (str, optional):               Index URL for Python Package Manager to use. Please do not use any indices you don't trust. The authentication provided in the URL should not have write-access. If you are not careful, these can cause [supply chain attack](https://en.wikipedia.org/wiki/Supply_chain_attack)
```

cognite/experimental/_api/functions.py
HMEiding: What about replacing
```
If you are not careful, these can cause supply chain attack
```
with the same warning we have in the Confluence docs:
```
Be aware of the intrinsic security risks using additional indices carry.
```
? We can still link to Wiki, though.

sarakao: @HMEiding I agree, we should stay consistent with the other docs, so I believe we should talk here about **dependency confusion attack** and link to the Cognite official doc.

PR #273, opened 2022-02-09
limit the geospatial tests to the domain testing

vincent-cognite:
No need to retest all that is tested in the main python sdk. Just keeping one of the tests for geospatial domain testing.


tuanng-cognite (approved):
Nice! :trophy:

PR #275, opened 2022-02-14
loosen up the regex version

andersfylling:
I also ran `poetry lock --no-update` to see if anything was out of sync.


henrist:
Oh. And your lockfile is no longer in sync, so it needs to be updated.

PR #277, opened 2022-02-14
SBS-4595: add initial support for putting rasters into feature raster properties

vincent-cognite:


tuanng-cognite (approved):
looking good

PR #280, opened 2022-02-15
Widen typing-extension version to allow v4

henrist:
https://github.com/python/typing/blob/master/typing_extensions/CHANGELOG

> Dropped support for Python versions 3.5 and older, including Python 2.7.

Should be fine for this library as far as I can see.


andersfylling:
nice, can you update the version as well? so we get a new release

henrist:
@andersfylling Done

andersfylling (approved):

henrist:
I just realized that I should add a changelog entry as well. I'll add a new PR for that.

PR #286, opened 2022-02-18
fix exception

andersfylling:
## Checklist

### DIFF
--- a/cognite/experimental/data_classes/vision.py
+++ b/cognite/experimental/data_classes/vision.py
@@ -53,8 +53,10 @@ class CreatedDetectAssetsInFilesJob(CogniteResource):
         self.items = items
 
     def dump(self, camel_case: bool = False) -> Dict[str, Any]:
-        assert camel_case
-        return resource_to_camel_case(self)
+        if camel_case:
+            return resource_to_camel_case(self)
+        else:
+            return super().dump(camel_case=camel_case)
 
     @classmethod
     def _load(cls, resource: Union[Dict, str], cognite_client=None):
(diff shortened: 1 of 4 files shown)


cognite/experimental/data_classes/vision.py
```diff
@@ -53,8 +53,10 @@ def __init__(
         self.items = items
 
     def dump(self, camel_case: bool = False) -> Dict[str, Any]:
-        assert camel_case
-        return resource_to_camel_case(self)
+        if camel_case:
+            return resource_to_camel_case(self)
+        else:
+            return super().dump(camel_case=camel_case)
```
torbjornopheim, line 59: Shouldn't this be the other way around?

torbjornopheim (approved):
LGTM

cognite/experimental/data_classes/vision.py
andersfylling: Yeah, it's a bit awkward here. The resource_to_camel_case is needed to deal with recursive conversion.

torbjornopheim:
Can you point to where the string convertion of the cognite resource happens in the code? So it's clearer what needs to be patched in the base sdk? @andersfylling

andersfylling:
@torbjornopheim I believe this is the issue: https://github.com/cognitedata/cognite-sdk-python/blob/26b24e1c34675c3e541bc7771e3bebf24b3a232d/cognite/client/data_classes/_base.py#L13

My understanding is that the first line returns either a dict or list[dict], and the json dump method does not have logic to handle dict nor list of dict which causes this to fail.

andersfylling:
this seems to fix the main issue in the experimental sdk: https://github.com/cognitedata/cognite-sdk-python-experimental/pull/287

PR #289, opened 2022-02-26
Update entity matching pipelines doc

rebwib:
This PR solves https://cognitedata.atlassian.net/browse/CXT-563 and https://cognitedata.atlassian.net/browse/CXT-564 and does some smaller improvements to the entity matching pipelines documentation.


peetcremer1 (approved):
LGTM! Just curious: I was expecting some updates for Pipelines and the removal `schedule_interval` Why is the relationship_label removed as well?

rebwib:
> LGTM! Just curious: I was expecting some updates for Pipelines and the removal `schedule_interval` Why is the relationship_label removed as well?

All the logic for `schedule_interval` and `relationship_label` happens in the worker. Hence, I just remove the possibility to use these parameters in the sdk and then we can do the  tasks in https://cognitedata.atlassian.net/browse/CXT-561 that removes these from context-api and remove the logic from the worker when we have time. 
The `relationship_label` can be set to true and then the matches created by the pipeline will be stored as relationships. however, this is only possible for a few whitelisted projects and not really used. Thus, I think it is better to remove it as I think it will be confusing to the user.

PR #293, opened 2022-03-10
SBS-4823: url encode feature external id and raster id

vincent-cognite:
feature external id and raster id have "free" content so we need to url encode them to avoid illegal urls (giving 404)


tuanng-cognite (approved):
Great!

PR #304, opened 2022-04-22
VIS-789: Filter annotations by data

marvrez:
Adds the `data` field to AnnotationV2Filter in correspondance with the
recent changes to the AnnotationsV2 API that introduces support for
querying of the annotation data field.

## Checklist


peetcremer1 (changes requested):
Changes look good! I think we should extend the integration test to make use of the data field through the AnnotationFilter class.

peetcremer1 (approved):
LGTM

PR #307, opened 2022-05-06
Geospatial dataset

tuanng-cognite:
We are trying to support data_set_id field in FeatureType object for some certain methods. For example, creating feature types
https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/_api/geospatial.py#L29

The problem is that FeatureType class is defined in official sdk.
The approach here is to override the creating feature types method in experimental sdk and re-define the FeatureType class here.

I don't know if there is a better approach?


vincent-cognite:
@tuanng-cognite shouldn't we simply add the dataset id field in the main sdk?
What happens if we add this field in a cluster with dataset id disable? It should just remain undefined, right?

tuanng-cognite:
https://github.com/cognitedata/cognite-sdk-python/pull/923
We only need to bump the main sdk version in this repository

vincent-cognite:
> [cognitedata/cognite-sdk-python#923](https://github.com/cognitedata/cognite-sdk-python/pull/923) We only need to bump the main sdk version in this repository

but this is not what you are doing here 🤔

tuanng-cognite:
> > [cognitedata/cognite-sdk-python#923](https://github.com/cognitedata/cognite-sdk-python/pull/923) We only need to bump the main sdk version in this repository
> 
> but this is not what you are doing here thinking

It is just that I'm slow.

vincent-cognite:
> It is just that I'm slow.

no, you are not!

dmivankov (approved):

vincent-cognite (approved):

PR #308, opened 2022-05-09
Annotation suggest

peetcremer1:
## Checklist

- [x] Linking documentation changes
- [x] Changelog entry added in `CHANGELOG` 
- [x] Version updated in `pyproject.toml`


tests/tests_integration/test_contextualization/test_annotations_v2.py
```diff
@@ -114,6 +136,24 @@ def test_create_annotations(self, cognite_client: CogniteClient, base_annotation
...
+        suggested_annotation = cognite_client.annotations_v2.suggest(base_suggest_annotation)
+        assert isinstance(suggested_annotation, AnnotationV2)
+        check_created_vs_base(base_suggest_annotation, suggested_annotation)
+        assert suggested_annotation.creating_user == None
+
+    def test_suggest_annotations(self, cognite_client: CogniteClient, base_suggest_annotation: AnnotationV2) -> None:
+        suggested_annotations = cognite_client.annotations_v2.suggest([base_suggest_annotation] * 30)
+        assert isinstance(suggested_annotations, AnnotationV2List)
+        for a in suggested_annotations:
+            check_created_vs_base(base_suggest_annotation, a)
```
marvrez, line 151: Might be redudant, but should we also check that `a.creating_user == None` here as well?

marvrez (approved):
LGTM!

cognite/experimental/_api/annotations_v2.py
```diff
@@ -26,6 +27,34 @@ def create(self, annotations: Union[AnnotationV2, List[AnnotationV2]]) -> Union[
...
+        return self._create_multiple(resource_path=self._RESOURCE_PATH + "/suggest", items=items)
+
+    @staticmethod
+    def _sanitize_suggest_item(annotation: AnnotationV2) -> Dict[str, any]:
+        # Check that status is set to suggested if it is set and afterwards remove it
+        item = annotation.dump(camel_case=True) if isinstance(annotation, AnnotationV2) else deepcopy(annotation)
+        if "status" in item:
+            if item["status"] != "suggested":
+                raise ValueError("status field for Annotation suggestions must be set to 'suggested'")
+            del item["status"]
```
cyriaquebrousse, line 55: Whether we need to protect the user from their obvious mistakes or not is a bit up to you as the author of the PR. But my take on it is it's perfectly fine to let the request go through to the API, and let the API fail. In that way we also eliminate redundant logic here in the SDK.

peetcremer1: You are right and I was debating that actually. It is completely ok to fail. 
I decided in the end to make it explicitly fail because the user is forced to set the "status" field when creating an instance of the `AnnotationV2` dataclass and if they set it to "approved" or "rejected" they would get very confused if the created annotations silently end up as "suggested"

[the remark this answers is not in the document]
peetcremer1: We need to delete because the `/suggest` route at API expects a data structure which does not contain the `status` field (as it is always "suggested"). Not deleting it would lead to a `400 BadRequest` on the API side

PR #313, opened 2022-06-08
feat: Remove linked resource in Annotations_V2

greenbech:
## Checklist

Not able to run the tests locally—do I need some kind of API key to do that? If they are trigged by CI here, that's probably enough. Will add the changelog and update package after I see that the test run

### DIFF
--- a/cognite/experimental/data_classes/annotations_v2.py
+++ b/cognite/experimental/data_classes/annotations_v2.py
@@ -21,16 +21,12 @@ class AnnotationV2(CogniteResource):
         status (str): The status of the annotation, e.g. "suggested", "approved", "rejected".
 
         annotated_resource_type (str): Type name of the CDF resource that is annotated, e.g. "file".
-        annotated_resource_id (int, optional): The internal ID of the annotated resource. Either this field or the external ID of the linked resource is required (not both).
-        annotated_resource_external_id (str, optional): The external ID of the annotated resource. Either this field or the internal ID of the linked resource is required (not both).
+        annotated_resource_id (int, optional): The internal ID of the annotated resource. Either this field or `annotated_resource_external_id` is required (not both).
+        annotated_resource_external_id (str, optional): The external ID of the annotated resource. Either this field or `annotated_resource_id` is required (not both).
 
         creating_app (str): The name of the app from which this annotation was created.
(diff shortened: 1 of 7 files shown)


cognite/experimental/data_classes/annotations_v2.py
```diff
@@ -21,17 +21,13 @@ class AnnotationV2(CogniteResource):
         status (str): The status of the annotation, e.g. "suggested", "approved", "rejected".
 
         annotated_resource_type (str): Type name of the CDF resource that is annotated, e.g. "file".
-        annotated_resource_id (int, optional): The internal ID of the annotated resource. Either this field or the external ID of the linked resource is required (not both).
-        annotated_resource_external_id (str, optional): The external ID of the annotated resource. Either this field or the internal ID of the linked resource is required (not both).
+        annotated_resource_id (int, optional): The internal ID of the annotated resource. Either this field or `annotated_resource_external_id` is required (not both).
+        annotated_resource_external_id (str, optional): The external ID of the annotated resource. Either this field or `annotated_resource_id` is required (not both).
```
marvrez, line 25: We don't support `annotated_resource_external_id` anymore. So the latter part of the comments can be removed.

marvrez: See: https://github.com/cognitedata/context-api/blob/6c38e2d895726fc990f46471fb72040a2fbca738/src/api/annotations_v2/schemas.py#L56

greenbech: If it isn't supported should we just completely remove the field?

marvrez: We have to do that at some point, and the API will anyways return an error if we attempt to create an annotation with an external id. So I think we should just remove this field now

peetcremer1: Not yet, as we need to push through the migration that removes it for good. We need to still be able to process those annotations that have the field. However, I approve of removing the documentation for it. Should probably also add a #TODO on the field itself that it should be removed as soon as the migration that removes the field on the data persistence level is through.

peetcremer1 (approved):
LGTM

PR #320, opened 2022-06-27
[MLOPS-1089] Added parsing of requirements function docstring

simenkristoff:
## Checklist


qTipTip (approved):
This looks great to me! 🚀 

We need to add this feature to the documentation as well. 

Also, could we use the same method to also validate the users `requirements.txt` when deploying using the other deployment methods? (In a separate task of course).

simenkristoff:
> 

Sure, i'll make issues for both documentation and validation of folder-uploads

PR #321, opened 2022-06-29
[MLOPS-1160] Added validation of requirements in folder upload

simenkristoff:
## Checklist


qTipTip (approved):
Looks good to me! Good job with the tests! :) 

**Optional**: If you want, you can take a look at swapping out the `os.path.join`-code with `pathlib`-equivalents. Since around Python 3.4 Pathlib is the preferred way of working with filepaths, and we still have some remnants of the old `os.path` way of doing things.

https://miguendes.me/python-pathlib is a good list of Pathlib-patterns.

PR #323, opened 2022-07-20
Alerts API alpha version

polomani:
## Checklist

It lacks some decent documentation for now, but this is not crucial at the moment


cognite/experimental/_api/alerts.py
```diff
@@ -0,0 +1,258 @@
...
+        Returns:
+            AlertChannelList: list of channels"""
+
+        filter = AlertChannelFilter(
+            external_ids=external_ids,
+            ids=ids,
+            parent_ids=parent_ids,
+            metadata=metadata,
+        ).dump(camel_case=True)
+        filter = {to_camel_case(k): v for k, v in (filter or {}).items() if v is not None}
```
VerstraeteBert: Keys should already be camel cased, no? Furthermore, the filter object can't be None. Is the reason for performing this extra filtering the omission of keys with nil values?

```suggestion
        filter = {k: v for k, v in filter.items() if v is not None}
```

cognite/experimental/data_classes/alerts.py
```diff
@@ -0,0 +1,244 @@
...
+
+        def remove(self, value: List) -> "AlertChannelUpdate":
+            return self._remove(value)
+
+    @property
+    def external_id(self) -> "_PrimitiveAlertChannelUpdate":
+        return AlertChannelUpdate._PrimitiveAlertChannelUpdate(self, "externalId")
+
+    @property
+    def source(self) -> "_PrimitiveAlertChannelUpdate":
```
VerstraeteBert, line 75: I think "name" makes more sense as the name for this property. A channel does not necessarily need to be identified by a source, as it can have many. We can have this semantic discussion some time later ;)

cognite/experimental/data_classes/alerts.py
```diff
@@ -0,0 +1,244 @@
...
+    def __init__(
+        self,
+        external_id: str = None,
+        id: int = None,
+        name: str = None,
+        parent_id: int = None,
+        parent_external_id: str = None,
+        description: str = None,
+        metadata: Dict[str, str] = None,
+        cognite_client: "CogniteClient" = None,
```
VerstraeteBert, line 26: Is the cognite_client here really nullable or required? Same with the other endpoints. We're also possibly casting None to CogniteClient later on, which will mislead the type checkers.

Not sure if I'm understanding the CogniteResource base class fully ;)

VerstraeteBert (commented):
Looks good. Have a few questions before approval :)

cognite/experimental/data_classes/alerts.py
polomani: prob a question to authors, but I see this is how others are using it :)

cognite/experimental/data_classes/alerts.py
polomani: yes, it was a bad copy-paste. will fix

cognite/experimental/_api/alerts.py
polomani: fixed, removed

VerstraeteBert (approved):
Thanks for clarifying, LGTM!

PR #327, opened 2022-08-02
Doc: Add example to Vision's annotate method

miladcognite:
Also bump the version and update `CHANGELOG`

## Checklist

### DIFF
--- a/cognite/experimental/_api/vision.py
+++ b/cognite/experimental/_api/vision.py
@@ -87,15 +87,7 @@ class VisionAPI(ContextAPI):
             AnnotateJobResults: Resulting queued job, which can be used to retrieve the status of the job or the annotation results if the job is finished. Note that .result property of this job will wait for the job to finish and returns the results.
 
         Examples:
-            Start a job and get the status::
-
-                >>> from cognite.experimental import CogniteClient
-                >>> from cognite.experimental.data_classes.vision import Feature
-                >>> c = CogniteClient()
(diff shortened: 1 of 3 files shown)


cognite/experimental/_api/vision.py
```diff
@@ -85,6 +85,26 @@ def annotate(
...
             AnnotateJobResults: Resulting queued job, which can be used to retrieve the status of the job or the annotation results if the job is finished. Note that .result property of this job will wait for the job to finish and returns the results.
+
+        Examples:
+            Start a job and get the status::
+
+                >>> from cognite.experimental import CogniteClient
+                >>> from cognite.experimental.data_classes.vision import Feature
+                >>> c = CogniteClient()
+                >>> annotate_job = c.vision.annotate(features=Feature.ASSET_TAG_DETECTION, file_ids=[1])
+                >>> annotate_job.update_status()
```
marvrez, line 96: I think you can remove this line. We don't really want/need the user to explicitly poll the job. They can just use `annotate_job.wait_for_completion` or `annotate_job.items` and the SDK will do the actual job status polling behind the scenes.

marvrez (approved):
LGTM. Added one minor comment

cognite/experimental/_api/vision.py
miladcognite: [727a012](https://github.com/cognitedata/cognite-sdk-python-experimental/pull/327/commits/727a012c2d1c8ad27d53cd06072eccfbc491c1ff)

PR #334, opened 2022-08-22
Docs: Added quickstart guide for Vision extract

miladcognite:
## Checklist

### DIFF
--- a/docs/source/cognite.rst
+++ b/docs/source/cognite.rst
@@ -342,7 +342,7 @@ Wait for job completion and get the parsed results:
     ...     # do something with the predictions
 
 
-Save the prediction results in CDF as `Annotations <https://docs.cognite.com/api/v1/#tag/Annotations/operation/annotationsCreate>`_:
+Save the prediction results in CDF as `Annotations <https://docs.cognite.com/api/v1/#tag/Annotations>`_:
 
 .. code:: python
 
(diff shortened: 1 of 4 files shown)


docs/source/cognite.rst
```diff
@@ -316,7 +316,41 @@ Vision
...
+
+.. code:: python
+
+    >>> extract_job.wait_for_completion()
+    >>> for item in extract_job.items:
+    ...     predictions = item.predictions
+    ...     # do something with the predictions
+
+
+Save the prediction results in CDF as `Annotations <https://docs.cognite.com/api/v1/#tag/Annotations/operation/annotationsCreate>`_:
```
marvrez, line 345: Should we instead point to the main annotations page instead of `annotationsCreate`? We do anyways point to the latter below.
```suggestion
Save the prediction results in CDF as `Annotations <https://docs.cognite.com/api/v1/#tag/Annotations>`_:
```

marvrez (approved):
LGTM. Added one minor suggestion

marvrez:
Also, since we are bumping the package version, remember to update the CHANGELOG.

docs/source/cognite.rst
miladcognite: I have no strong preferences. Will add your suggestion

PR #341, opened 2022-09-04
SBS-5407: add upsert for geospatial

vincent-cognite:


vincent-cognite:
giving up trying the upgrade to latest cognite sdk

PR #348, opened 2022-09-15
add-alert-list-subscribers-with-filter

hamdielhamdi-cognite:
## Checklist

- [ ] Version updated in `pyproject.toml` or not relevant for this change

### DIFF
--- a/cognite/experimental/_api/alerts.py
+++ b/cognite/experimental/_api/alerts.py
@@ -12,6 +12,7 @@ from cognite.experimental.data_classes.alerts import (
     AlertFilter,
     AlertList,
     AlertSubscriber,
+    AlertSubscriberFilter,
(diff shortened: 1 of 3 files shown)


cognite/experimental/_api/alerts.py
```diff
@@ -119,6 +120,34 @@ def create(
...
+            AlertSubscriberList: list of subscribers"""
+
+        filter = AlertSubscriberFilter(
+            external_ids=external_ids,
+            ids=ids,
+            email=email,
+            metadata=metadata,
+        ).dump(camel_case=True)
+
+        return self._list(method="POST", limit=limit, filter=filter)
```
polomani, line 149: does this work? limit is not a valid property (not yet implemented)

PR #357, opened 2022-12-13
SBS-5698: add support for geospatial feature type partitions

vincent-cognite:


cognite/experimental/_api/geospatial.py
```diff
@@ -52,6 +59,101 @@ def __init__(self, *args, **kwargs):
                 wrapped = _with_cognite_domain(getattr(GeospatialAPI, attr_name))
                 setattr(ExperimentalGeospatialAPI, attr_name, wrapped)
 
+    @_with_cognite_domain
+    def create_feature_types(
```
vincent-cognite, line 63: in order to use the new definition with partitions for FeatureType, I don't see another way to repeat the implementation for create_feature_types(), list_feature_types() and retrieve_feature_types()
Do you know a better approach @tuanng-cognite ?

tuanng-cognite: I don't see any other ways either. The point is that the create_feature_types changes, receives another input object.

cognite/experimental/_api/geospatial.py
```diff
@@ -52,6 +59,101 @@ def __init__(self, *args, **kwargs):
...
+
+    @overload
+    def retrieve_feature_types(self, external_id: str) -> FeatureType:
+        ...
+
+    @overload
+    def retrieve_feature_types(self, external_id: List[str]) -> FeatureTypeList:
+        ...
+
+    def retrieve_feature_types(self, external_id: Union[str, List[str]]) -> Union[FeatureType, FeatureTypeList]:
```
tuanng-cognite, line 131: with_domain annotation here?

cognite/experimental/_api/geospatial.py
vincent-cognite: if I keep using the "main" implementations for list_feature_types() and retrieve_feature_types(), I still get the `partitions` field, so I don't need to reimplement them.
Only necessary for create_feature_types() afais

tuanng-cognite (approved):

PR #364, opened 2023-04-27
lock cognite-sdk version

VerstraeteBert:
Update and lock cognite-python-sdk version. Since this package depends on internal functions, e.g., stuff in  `cognite.client.utils._auxiliary`, it should be strictly locked. Have observed pulling this library, with a newer version of cognite-sdk than specified in the poetry.lock (but still allowed by pyproject) and getting errors because of moving / removal of these internal functions. 

Note, we should really start using a stricter linting/type-checking setup, I see a ton of unused imports, typing mismatches in my IDE. Maybe go with ruff and pyright @cognitedata/matchmakers ?

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,14 @@ Changes are grouped as follows
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## [0.112.0]
+
+### Changed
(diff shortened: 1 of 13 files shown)


polomani (approved):
approving to unblock :)

olacognite:
Locking version makes sense to me

PR #366, opened 2023-05-05
Lock cognite-sdk version to major

polomani:
## Checklist
need to relax the constraint as people cannot run this on newer versions of cognite-sdk

- [x ] Changelog entry added in `CHANGELOG` or not relevant for this change
- [ x] Version updated in `pyproject.toml` or not relevant for this change


evertoncolling (approved):
LGTM 👍

PR #367, opened 2023-05-09
Update regex dependency

evertoncolling:
## Checklist

Updated the `regex` dependency to enable using the `cognite-sdk-python-experimental` in Pyodide.


pyproject.toml
```diff
@@ -49,7 +49,7 @@ responses = "^0.13.3"
 sympy = "^1.3.0"
 typing-extensions = ">=3.7.4,<5"
 pandas = ">=1.1.5"
-regex = ">=2020.11.13, <2023"
+regex = ">=2020.11.13"
```
polomani, line 52: this is quite unsafe, I think we should review it at least sometimes.
how about 2024?

evertoncolling: Agree! I wasn't sure how this way of versioning worked, but I guess setting the upper boundary to `2024` should do the trick.

polomani:
@evertoncolling CI seems to be hanging

PR #370, opened 2023-08-09
Add simulator runs docs

polomani:
- added simulation runs api docs
- removed deprecated model hosting docs

Screenshot 2023-08-09 at 16 36 58

it will appear here: https://cognite-sdk-experimental.readthedocs-hosted.com/en/latest/cognite.html#cogniteclient


evertoncolling (approved):
👍

PR #372, opened 2023-09-21
Deprecate legacy annotations

cyriaquebrousse:
We have announced that legacy annotations would be removed this spring, so now we're making this official in the SDKs.
"New" annotations have been available since late 2022.

## Checklist


olacognite (approved):
LGTM!

PR #373, opened 2023-09-21
Remove duplicated v1 Annotation methods

cyriaquebrousse:
The methods provided in `annotations.py` are duplicated from the [main SDK](https://cognite-sdk-python.readthedocs-hosted.com/en/latest/data_organization.html#annotations), so we are removing them from the experimental SDK.

Since the functionality is available there exactly as-is, there is no need for a deprecation window.

Closes CXT-1385.

## Checklist


olacognite (approved):
LGTM! Given fixing imports so that tests run again though

